PR #3, opened 2021-03-30
Check in 104.py.

woke7:


mujingw (commented):
104 is the maximum depth problem. You committed a wrong file.

woke7:
> 104 is the maximum depth problem. You committed a wrong file.

Fixed.

mujingw (approved):
LGTM. This is a piece of very well structured code. Line 9 handles the base case where the node is `None`. Line 12 handles another base case where the node is a leaf node (neither left nor right child exists). Line 15 is the recursive step.

PR #4, opened 2021-03-30
Check in 111.py.

woke7:


mujingw (commented):
LGTM. However, since this is a minimum distance problem, I would use a BFS approach, which usually finds the answer quicker than a DFS approach as shown in your code (basically, BFS works in a fail-fast/return-early style for this kind of problems). 

Personally I always find it hard to understand and write DFS code correctly, even for a simple problem like this one. While in BFS, I have much more confidence in my solution.

```python

def minDepth(self, root):
    if not root:
        return 0
    
    q = collections.deque([(root, 1)])
    
    while q:
        curr, dist = q.popleft()
        
        if not curr.left and not curr.right:
            return dist
        
        if curr.left:
            q.append((curr.left, dist + 1))
        
        if curr.right:
            q.append((curr.right, dist + 1))

woke7:
> LGTM. However, since this is a minimum distance problem, I would use a BFS approach, which usually finds the answer quicker than a DFS approach as shown in your code (basically, BFS works in a fail-fast/return-early style for this kind of problems).
> 
> Personally I always find it hard to understand and write DFS code correctly, even for a simple problem like this one. While in BFS, I have much more confidence in my solution.
> 
> ```python
> def minDepth(self, root):
>     if not root:
>         return 0
>     
>     q = collections.deque([(root, 1)])
>     
>     while q:
>         curr, dist = q.popleft()
>         
>         if not curr.left and not curr.right:
>             return dist
>         
>         if curr.left:
>             q.append((curr.left, dist + 1))
>         
>         if curr.right:
>             q.append((curr.right, dist + 1)) 
> ```



> LGTM. However, since this is a minimum distance problem, I would use a BFS approach, which usually finds the answer quicker than a DFS approach as shown in your code (basically, BFS works in a fail-fast/return-early style for this kind of problems).
> 
> Personally I always find it hard to understand and write DFS code correctly, even for a simple problem like this one. While in BFS, I have much more confidence in my solution.
> 
> ```python
> def minDepth(self, root):
>     if not root:
>         return 0
>     
>     q = collections.deque([(root, 1)])
>     
>     while q:
>         curr, dist = q.popleft()
>         
>         if not curr.left and not curr.right:
>             return dist
>         
>         if curr.left:
>             q.append((curr.left, dist + 1))
>         
>         if curr.right:
>             q.append((curr.right, dist + 1)) 
> ```

Got it, make sense. I will use BFS next time for a similar problem.

mujingw:
Even though this commit is now merged, line 15 and line 18 don't really make sense. The recursive function already defined `None` node as a base case, why are we handling `None` children (as in line 15 and line 18) in the recursive steps again? We should delegate the handling of `None` nodes to the next level of the recursive call, and only keep the absolute minimum steps required in the recursive function body. The logic should be structured in a way similar to the solution for problem #4 . 

When designing recursive functions, the current level should only take care of itself, not the level above or the level below it. We always assume:
1. whatever returned **to** the current level from those lower level recursive calls is a correct answer of the sub-problem (i.e., we know the answers of the smaller problems)
2. whatever result returned to the caller **from** the current level is an aggregated answer that reflects the situation of both the current level and the levels below the current level

Therefore, the only task for each recursive step is to construct an answer based on sub-problem answers and the information available from the current level.  

This is a very important point when we face more complicated topics related to recursion, such as backtracking and top-down dynamic programming. Our code will become very messy if we fail to do so.

PR #5, opened 2021-03-31
Check in 141.py.

woke7:


mujingw (approved):
LGTM. Line 9 is very good.

PR #8, opened 2021-03-31
Check in 796.py.

woke7:


mujingw (commented):
For string (or array) rotation problems, a good technique to apply is to create a new string (or array) by concatenating two of the original strings (or arrays) and then do our work. Your code is still very much in a C-style with multiple handlings on boundary cases. 

For this particular problem, all we need to do is just to check whether or not `B` is in `A + A`. That's it!

```python
return B in A + A and len(B) == len(A)

File: 796.py [hunk 1]
```diff
@@ -1,18 +1,3 @@
...
-        
-        while length > 0:
-            length -= 1
-            A = A[1:len(A)] + A[0:1]
-            
-            if A == B:
-                return True
-        
-        return False
+        return len(A) == len(B) and B in A + A
```
mujingw, line 3: LGTM. Checking whether or not `B` is in `A + A` only takes linear time `O(length)`. In your previous implementation, you checked `A == B` for `length` times, and each check takes `O(length)` time, making the overall complexity `O(length^2)`, which was not good.

mujingw (approved):

796.py
```diff
@@ -0,0 +1,18 @@
...
+            return False
+        
+        if len(A) == 0 and len(B) == 0:
+            return True
+        
+        length = len(A)
+        
+        while length > 0:
+            length -= 1
+            A = A[1:len(A)] + A[0:1]
```
mujingw: This part is expensive too. It takes linear time to construct a new string that appends `A[0]` to the end of `A[1:]`.

Also, a pythonic way of writing `A[1:len(A)]` is `A[1:]`.

796.py
```diff
@@ -0,0 +1,18 @@
...
+            
+            if A == B:
```
mujingw: For the record, it is expensive to do this operation in every single iteration.

PR #10, opened 2021-04-01
Check in 832.py.

woke7:


832.py
```diff
@@ -0,0 +1,12 @@
+from typing import List
+
+class Solution:
+    def flipAndInvertImage(self, image: List[List[int]]) -> List[List[int]]:
```
mujingw, line 4: Similar to the other problem, we should allocate new space for the resulting image, not modifying the input `image`. If the problem asked you to do everything in-place and not to return anything, then your implementation would be good.

832.py
```diff
@@ -3,10 +3,10 @@
 class Solution:
     def flipAndInvertImage(self, image: List[List[int]]) -> List[List[int]]:
         # Flip the image horizontally.
-        image = [row[::-1] for row in image]
+        flipped_image = [row[::-1] for row in image]
```
mujingw, line 6: LGTM. This is pythonic.

mujingw (approved):

mujingw (approved):

PR #12, opened 2021-04-01
Check in 344.py.

woke7:


344.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Solution:
+    def reverseString(self, s: List[str]) -> None:
+        """
+        Do not return anything, modify s in-place instead.
+        """
+        start = 0
+        end = len(s) - 1
+        
+        while start < end:
```
mujingw, line 11: Not pythonic. We do swaps in Python like this: 
```python
a, b = b, a

mujingw (approved):
LGTM

PR #13, opened 2021-04-01
Check in 1295.py.

woke7:


1295.py
```diff
@@ -0,0 +1,20 @@
+from typing import List
+
+class Solution:
+    def findNumbers(self, nums: List[int]) -> int:
+        count = 0
+        
+        for num in nums:
+            if self.computeDigitNumer(num) % 2 == 0:
```
mujingw, line 8: Good modularization!

1295.py
```diff
@@ -0,0 +1,20 @@
...
+                count += 1
+        
+        return count
+
+    def computeDigitNumer(self, num: int) -> int:
```
mujingw, line 13: Typo here. Also, the function name can be simplified. How about `countDigits`?

mujingw (approved):
LGTM

PR #14, opened 2021-04-01
Check in 804.py.

woke7:


804.py
```diff
@@ -0,0 +1,11 @@
+from typing import List
+
+class Solution:
+    def uniqueMorseRepresentations(self, words: List[str]) -> int:
+        trans = set()   
+        morse = [".-","-...","-.-.","-..",".","..-.","--.","....","..",".---","-.-",".-..","--","-.","---",".--.","--.-",".-.","...","-","..-","...-",".--","-..-","-.--","--.."]
```
mujingw, line 6: It's [recommended](https://www.python.org/dev/peps/pep-0008/#constants) that constants are written in all capital letters. `morse` here is a constant variable in your program.

mujingw (approved):
LGTM

PR #15, opened 2021-04-01
Check in 56.py.

woke7:


56.py
```diff
@@ -0,0 +1,22 @@
+from typing import List
+
+class Solution:
+    def merge(self, intervals: List[List[int]]) -> List[List[int]]:
+        intervals.sort(key=self.compare)
```
mujingw, line 5: Two issues:

1. This modifies the original list, which is not good. Use `my_sorted_list = sorted(input_list)` instead. Also, check the documentation and learn the difference between `sorted(some_list)` and `some_list.sort()`.
2. In Python, we usually use a lambda expression to specify the comparator, especially when the sort logic is simple like this one. 
```python
sorted_intervals = sorted(intervals, key=lambda x: x[0])

File: 56.py [hunk 2]
```diff
@@ -0,0 +1,22 @@
+from typing import List
+
+class Solution:
+    def merge(self, intervals: List[List[int]]) -> List[List[int]]:
+        intervals.sort(key=self.compare)        
+        merged = []
+        current = intervals[0]
+        
+        for incoming in intervals[1:]:
```
mujingw, line 9: Should give names to `incoming[0]`, `incoming[1]`, `current[0]`, and `current[1]` for readability. For example,

```python
for incoming_start, incoming_end in intervals[1:]

Comment (woke7) [reply 56.py]: How do you know the retrieved list only has two elements in it?

Comment (mujingw) [reply 56.py]: What you get each time by iterating through `intervals[1:]` is not a list; what you get is an element from `intervals[1:]`, and each element in the `intervals` list has two fields: `intervals[i][0]` and `intervals[i][1]`. Here I am naming `interval[i][0]` and `interval[i][1]` as `incoming_start` and `incoming_end`, respectively.

File: 56.py (hunk 1 above)
Comment (mujingw) [reply 56.py]: Formally, lambdas are called anonymous functions, which comes in handy when a function definition is short and doesn't really need a function name due to the function's simplicity and its limited scope (no need to be visible by anyone else). It is a support provided by the programming language to simplify coding for cases like this one (as you can see, your original `compare` function consists of only one line of code).

File: 56.py [hunk 3]
```diff
@@ -2,21 +2,20 @@
...
-        intervals.sort(key=self.compare)        
+        sorted_intervals = sorted(intervals, key=lambda x : x[0])       
         merged = []
-        current = intervals[0]
+        current = sorted_intervals[0]
         
-        for incoming in intervals[1:]:
-            if incoming[0] > current[1]:
+        for incoming_start, incoming_end in sorted_intervals[1:]:
+            current_start = current[0]
```
mujingw, line 10: This is good because it makes line 15 and 17 more readable.

mujingw (approved):
LGTM

PR #17, opened 2021-04-02
Check in 102.py.

woke7:


mujingw (commented):
This implementation is alright. However, as you pointed out, you have to maintain two variables `prev` and `curr` in this implementation. More variables means:

1. More chances to make mistakes when implementing the idea
2. More fragmented structure in the function body, which might lead to maintenance problems down the road

Also, having to check the value of `prev` in every iteration is not ideal from a performance perspective.

mujingw (approved):
LGTM. This should be the standard of any code we write--no extraneous variables or blocks, and gets the task done efficiently.

PR #18, opened 2021-04-02
Check in 429.py.

woke7:


429.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Solution:
+    def levelOrder(self, root: 'Node') -> List[List[int]]:
+        levels = []
+        
+        if not root:
+            return levels
+        
+        queue = deque([root])
+        prev, curr, level = 1, 0, []
```
mujingw, line 20: This line is pythonic!

429.py
```diff
@@ -0,0 +1,38 @@
...
+        
+        while queue:
```
mujingw, line 22: Try to re-write this using the template I described [here](https://github.com/woke7/leetcode-python/pull/19#discussion_r606710471).

mujingw (approved):
LGTM

PR #19, opened 2021-04-02
Check in 107.py.

woke7:


107.py
```diff
@@ -0,0 +1,39 @@
...
+        if not root:
+            return levels
+        
+        queue = deque([root])
+        
+        while queue:
+            level = []
+            level_length = len(queue)
+            
+            while level_length > 0:
```
mujingw, line 24: This BFS implementation is very close to the BFS template I always use. However, instead of maintaining `level_length` yourself, why don't we just use a `for` loop, especially when `level_length` is a variable that has to be decremented every time regardless of what happens in the loop? This is a good opportunity to think about which scenarios are better suited for a `while` loop and which are for a `for` loop.

mujingw (approved):
LGTM

PR #21, opened 2021-04-02
Check in 103.py.

woke7:


103.py
```diff
@@ -0,0 +1,44 @@
...
+                    queue.append(node.right)
+                    
+                level_length -= 1
+            
+            if (toggle):
+                levels.append(reversed(level))
+            else:
+                levels.append(level)
+        
+            level = []
```
mujingw, line 41: `level` gets assigned a new empty list in 21. There is no need to reset it here.

103.py
```diff
@@ -0,0 +1,44 @@
...
+            toggle = not toggle
```
mujingw, line 42: This line is very smart.

103.py
```diff
@@ -0,0 +1,44 @@
...
+        levels = []
+        
+        if not root:
+            return levels
+        
+        queue = deque([root])
+        toggle = False
+        
+        while queue:
+            level = []
```
mujingw, line 21: As you can see, using this template we can make `level` local to the `while queue:` loop, whereas in your [previous template](https://github.com/woke7/leetcode-python/pull/17/commits/15d960af88781ca13ede8aaf9ddd41a820c7db15) you had to declare `level` outside of the loop. This is a good example of making your code less fragmented.

mujingw (approved):
LGTM

PR #22, opened 2021-04-02
Check in 199.py.

woke7:


199.py
```diff
@@ -0,0 +1,36 @@
...
+                if level_length == 1:
+                    right_nodes.append(node.val)
+                    
+                if node.left:
+                    queue.append(node.left)
+                    
+                if node.right:
+                    queue.append(node.right)
+                    
+                level_length -= 1
```
mujingw, line 34: Same as [this](https://github.com/woke7/leetcode-python/pull/19#discussion_r606710471)

woke7: This one is a bit different, as I need `level_length` in each iteration to determine if that the rightmost node at that level.

mujingw: You can still use a `for` loop:

```python
for i in range(level_length):
    if i == level_length - 1:
        # append your right-most node's value here
```

What's good about this `for` loop is that it avoids the manual maintenance of `level_length`.

PR #24, opened 2021-04-02
Check in 75.py.

woke7:


75.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+
+class Solution:
+    def sortColors(self, nums: List[int]) -> None:
+        """
+        Do not return anything, modify nums in-place instead.
+        """
+        start, mid, end = 0, 0, len(nums) - 1
```
mujingw, line 8: `mid` is a little misleading here. It's not necessarily the middle point between `start` and `end`, as in the binary search algorithm. What `mid` does is more like a sentinel that walks through each element in the array. Therefore, I think naming it `current` or simply `p` is more accurate and avoids confusion.  

On a separate note, this problem can be solved using the Counting Sort idea too.

mujingw (approved):
LGTM

PR #25, opened 2021-04-02
Check in 283.py.

woke7:


283.py
```diff
@@ -0,0 +1,13 @@
+from typing import List
+
+class Solution:
+    def moveZeroes(self, nums: List[int]) -> None:
+        """
+        Do not return anything, modify nums in-place instead.
+        """
+        start = 0
```
mujingw, line 8: `start` is misleading (although you could argue that it marks the start of the zero portion of the array). What it really does in the program is that it points to the first zero element in the array and marks the end of the non-zero portion. I would name it something like `end`, `boundary`, or `current` (current, because that's where you **currently** do the element swap operation).

mujingw (approved):
LGTM

PR #26, opened 2021-04-03
Check in 345.py.

woke7:


345.py
```diff
@@ -0,0 +1,20 @@
+class Solution:
+    def reverseVowels(self, s: str) -> str:
+        start, end = 0, len(s) - 1
+        vowels = set(['a', 'e', 'i', 'o', 'u', 'A', 'E', 'I', 'O', 'U'])
+        word = list(s)
+        
+        while start < end:
```
mujingw, line 7: Not bad, but we can still make a few improvements. 

1. Loop invariant `start` < `end` has to be true every time we enter the loop and has to remain true whenever we are still in the loop.
2. Based on 1, can we search for the first vowel from the left hand side faster?
3. Based on 1, can we search for the first vowel from the right hand side faster?
4. Once we find both, we can now do a swap.
5. After the swap is done, the problem that remains to be solved becomes smaller than the original problem, because all we care about now is the content within `word[start+1: end]` (left boundary inclusive and right boundary exclusive).

Based on the description above, can you simplify your logic inside the loop?

mujingw (approved):
Great! This implementation is exactly what I expected.

PR #28, opened 2021-04-03
Check in 771.py.

woke7:


771.py
```diff
@@ -0,0 +1,10 @@
+class Solution:
+    def numJewelsInStones(self, jewels: str, stones: str) -> int:
+        jewels_set = set(list(jewels))
```
mujingw, line 3: Redundant step by converting the string to a list and then convert to a set. You can do it in one step as mentioned [here](https://github.com/woke7/leetcode-python/pull/26#discussion_r606714633).

771.py
```diff
@@ -0,0 +1,10 @@
...
+        jewels_num = 0
```
mujingw, line 4: `jewels_count` would be more accurate. There is a slight difference between a count and a number.

PR #29, opened 2021-04-03
Check in 1122.py.

woke7:


1122.py
```diff
@@ -0,0 +1,17 @@
+from collections import defaultdict
+from typing import List
+
+class Solution:
+    def relativeSortArray(self, arr1: List[int], arr2: List[int]) -> List[int]:
+        only_arr1 = sorted([e for e in arr1 if e not in arr2])
+        num_pos = defaultdict(lambda: 0)
```
mujingw, line 7: For common default values like an `int`, you can just say `defaultdict(int)`, as mentioned [here](https://github.com/woke7/leetcode-python/commit/3f000781c8b70971b26b988834ab84606fdff4f3#r48915408).

1122.py
```diff
@@ -0,0 +1,17 @@
+from collections import defaultdict
+from typing import List
+
+class Solution:
+    def relativeSortArray(self, arr1: List[int], arr2: List[int]) -> List[int]:
+        only_arr1 = sorted([e for e in arr1 if e not in arr2])
```
mujingw, line 6: `arr2` is an array; therefore, it is quite expensive to check if every `e` in `arr1` is not in `arr2`. Each individual search takes O(N) time, and you did it for all N elements in `arr1`.

1122.py
```diff
@@ -3,8 +3,10 @@
 
 class Solution:
     def relativeSortArray(self, arr1: List[int], arr2: List[int]) -> List[int]:
-        only_arr1 = sorted([e for e in arr1 if e not in arr2])
-        num_pos = defaultdict(lambda: 0)
+        arr1_set = set(arr1)
```
mujingw, line 6: `arr1_set` is not needed. There isn't any **look-up** operation related to elements in `arr1` in the rest of the program. In line 8, you will go through each element in `arr1` anyway.

PR #30, opened 2021-04-03
Check in 136.py.

woke7:


136.py
```diff
@@ -0,0 +1,10 @@
+from typing import List
+
+class Solution:
+    def singleNumber(self, nums: List[int]) -> int:
+        single = 0
+
+        for num in nums:
```
mujingw, line 7: Your implementation is correct, but I would rather initialize `single` with the first element in `nums` and loop through `nums[1:]`.

PR #32, opened 2021-04-03
Check in 993.py.

woke7:


993.py
```diff
@@ -0,0 +1,38 @@
...
+                
+                if not node.left or not node.right:
+                    continue
+                
+                # Siblings
+                if (x == node.left.val and y == node.right.val) or \
+                   (x == node.right.val and y == node.left.val):
+                    return False
+
+            if x in level and y in level:
```
mujingw, line 35: 1. The time complexity of `x in level` can be decreased significantly if you choose your data structure more carefully. 
2. Other than your current implementation, can you think of a way of storing all the parent-child relationships, such that the cousin-checking logic can be simplified?

woke7: 1. Addressed.
2. Hmm...I can think of a way to store the parent-child relationship, but not sure how can I make the cousin-checking logic simpler with that though.

mujingw: Cousins by definition are people in the same generation but do not share the same parent. Your current logic is based on "generation" (or `level`, in the context of this problem) plus a lengthy check on whether or not `x` and `y` are siblings. If you maintain a data structure that enables you for a fast lookup on parent-child relationship, would you be able to simply your code between line 28 and 34?

woke7: Hmm...I cannot think of a better way, the comparison for checking siblings is pretty simple O(1) check, isn't it?

mujingw: It's more about simplifying the logic than decreasing the time complexity in this case. I was suggesting something like this:

```python
        q = collections.deque([(root, root)])
        
        while q:              
            size = len(q)
            level = {}
            
            for i in range(size):
                curr_node, curr_parent = q.popleft()
                level[curr_node.val] = curr_parent
                
                if curr_node.left:
                    q.append((curr_node.left, curr_node))
                if curr_node.right:
                    q.append((curr_node.right, curr_node))
                    
            if x in level and y in level:
                return level[x] != level[y]
        
        return False
```

For each node, you just need to save one more piece of information into `level` (i.e., the parent), and then both the siblings and the cousins scenarios can be handled at one single place.

PR #33, opened 2021-04-03
Check in 1512.py.

woke7:


1512.py
```diff
@@ -0,0 +1,14 @@
...
+        total_pairs = 0
+        
+        for count in Counter(nums).values():
+            if count > 1:
+                total_pairs += self.countPair(count)
+        
+        return total_pairs
+    
+    def countPair(self, count: int) -> int:
+        return count * (count - 1) // 2
```
mujingw, line 14: Good job! N choose 2. Or you can just use `comb(count, 2)`.

On a side note, using `Counter` to count the occurrences of all the numbers and then doing some other counting work is a common pattern of problems on LeetCode, e.g., [1679](https://leetcode.com/problems/max-number-of-k-sum-pairs/) and [1711](https://leetcode.com/problems/count-good-meals/).

mujingw: Today's bi-weekly contest also featured [a problem](https://leetcode.com/problems/count-nice-pairs-in-an-array/) similar to this one.

woke7: Will do as of today's 3 problems.

PR #34, opened 2021-04-04
Check in 1816.py.

woke7:


1816.py
```diff
@@ -0,0 +1,3 @@
+class Solution:
+    def truncateSentence(self, s: str, k: int) -> str:
+        return ' '.join(s.split(" ")[0:k])
```
mujingw, line 3: Two redundancies: 

1. No need to specify the `' '` character in `split()` when you split base on empty spaces.
2. If you slice the array starting from the 0 index, you don't need to specify the `0` either--just use `[:k]`.

PR #35, opened 2021-04-04
Check in 1817.py.

woke7:


1817.py
```diff
@@ -0,0 +1,12 @@
+class Solution:
+    def findingUsersActiveMinutes(self, logs: List[List[int]], k: int) -> List[int]:
+        user_to_mins = defaultdict(set)
```
mujingw, line 3: Very good. I used a dictionary of dictionaries during the contest. The inner dictionary was an overkill and I should have used a `set` like here.

mujingw (approved):
LGTM

PR #36, opened 2021-04-04
Check in 1812.py.

woke7:


mujingw (approved):
Very good. Simpler than mine.

PR #37, opened 2021-04-05
Check in 1470.py.

woke7:


1470.py
```diff
@@ -0,0 +1,11 @@
+class Solution:
+    def shuffle(self, nums: List[int], n: int) -> List[int]:
+        res = []
+        pointer = n
```
mujingw, line 4: `pointer` is a redundant variable. In each iteration, you can always append elements to `res` by accessing indexes `i` and `i + n` for `i` from `0` to `n-1`.

1470.py
```diff
@@ -1,11 +1,9 @@
 class Solution:
     def shuffle(self, nums: List[int], n: int) -> List[int]:
         res = []
-        pointer = n
         
-        for num in nums[:n]:
-            res.append(num)
-            res.append(nums[pointer])
-            pointer += 1
+        for i in range(len(nums[:n])):
```
mujingw, line 5: Isn't `len(nums[:n])` exactly the same as `n`?

```python
def shuffle(self, nums: List[int], n: int) -> List[int]:
    res = []
    
    for i in range(n):
        res.append(nums[i])
        res.append(nums[i + n])
        
    return res

PR #38, opened 2021-04-05
Check in 1365.py.

woke7:


1365.py
```diff
@@ -0,0 +1,16 @@
...
+        
+        for num in nums:
+            count = 0
+            
+            for sorted_num in sorted_nums:
+                if num > sorted_num:
+                    count += 1
+            
+            ans.append(count)
+            count = 0
```
mujingw, line 14: Same issue as [this](https://github.com/woke7/leetcode-python/pull/21#discussion_r606711085)

1365.py
```diff
@@ -0,0 +1,16 @@
+class Solution:
+    def smallerNumbersThanCurrent(self, nums: List[int]) -> List[int]:
+        sorted_nums = sorted(nums)
+        ans = []
+        
+        for num in nums:
+            count = 0
+            
+            for sorted_num in sorted_nums:
```
mujingw, line 9: `x = sorted_nums.index(num)` does the same thing as your inner `for` loop. However, both the `index()` function and your `for` loop take linear time to complete this task for every `num` in `nums`, making the overall time complexity `O(N^2)`. 

Since `sorted_nums` is already in a sorted order, can you think of a better way to locate the `x` mentioned above?

mujingw (commented):
You are still not taking the full advantage of a sorted array. When an array is sorted and you want to find a certain value in this array, binary search should immediately pop up in your mind. 

```python 
def smallerNumbersThanCurrent(self, nums: List[int]) -> List[int]:
    lookup = sorted(nums)
    res = []
    
    for num in nums:
        res.append(bisect.bisect_left(lookup, num))    
    
    return res
```

What `bisect_left` does is finding the **left-most** position for you to insert value `num` into the sorted array `lookup` while maintaining the sorted order even after the insertion. Here, you won't actually insert the value, but this function essentially helps you identify the number of values that are smaller than `num` in the sorted array `lookup` due to the "left most" guarantee: 

- In any array (whether it's sorted or not), a position's index always equals to the number of elements to its left. 
- In a sorted array, elements to the left of the left most insert position (as returned by `bisect_left`) all satisfy the condition `element < lookup[left_most_insert_position]`. 

For example, assuming that we have array `nums = [100, 101, 200, 200, 200, 303]`, `bisect_left(nums, 200)` will return you `2`, meaning that you would have to insert your new value `200` at index 2 (and right shift all the elements on and after index 2), making the new array `[100, 101, 200, 200, 200, 200, 303]`. The `2` returned to you by `bisect_left` is exactly the number of elements that are less than `200`--the value you "attempted" to insert. 

Note that `bisect_left` makes an assumption that the array you pass in is already sorted and is implemented using a binary search approach. Therefore, the time complexity of this function is `O(log(N))`.

1365.py
```diff
@@ -4,13 +4,14 @@ def smallerNumbersThanCurrent(self, nums: List[int]) -> List[int]:
         ans = []
         
         for num in nums:
-            count = 0
+            idx = sorted_nums.index(num)
```
mujingw, line 7: This is still `O(N)` complexity, as I mentioned in my [previous comment](https://github.com/woke7/leetcode-python/pull/38#discussion_r606906423). I specifically said that `index()` takes `O(N)` time.

1365.py
```diff
@@ -4,13 +4,14 @@ def smallerNumbersThanCurrent(self, nums: List[int]) -> List[int]:
...
             
-            for sorted_num in sorted_nums:
-                if num > sorted_num:
-                    count += 1
-            
-            ans.append(count)
-            count = 0
-        
+            while idx > 0:
```
mujingw, line 9: As specified in in the [docs](https://docs.python.org/3/tutorial/datastructures.html), `index()` returns you a `zero-based index in the list of the first item whose value is equal to x.` Therefore, there is no need to have this `while` loop shift your `idx` to the left any further. 

However, note that the implementation of `index()` makes no assumption on the list (or array) on whether or not it is already sorted, therefore it uses a linear searching algorithm to find the left-most position for you. This is the key difference between `list.index()` and `bisect.bisect_left()`.

mujingw: I need to give credit where credit is due: your `while` loop is correct, and it won't cause a `list index out of range` error. My point is that this `while` loop is trying to obtain an outcome that is already guaranteed by the `index()` function.

1365.py
```diff
@@ -1,17 +1,11 @@
...
-            idx = sorted_nums.index(num)
-            
-            while idx > 0:
-                if num > sorted_nums[idx - 1]:
-                    break
-                else:
-                    idx -= 1
-
-            ans.append(idx)
+            ans.append(bisect.bisect_left(sorted_nums, num))
```
mujingw, line 9: LGTM.

Just for the record, the following is what I would write if I didn't know `bisect.bisect_left()`:

```python 
def smallerNumbersThanCurrent(self, nums: List[int]) -> List[int]:
    lookup = sorted(nums)
    res = []
    
    for num in nums:
        res.append(lookup.index(num))    
    
    return res
```

It is slower, but the code is compact and correct.

mujingw (approved):

PR #39, opened 2021-04-05
Check in 1342.py.

woke7:


mujingw (approved):
LGTM. Instead of using `%`, `-`, and `/`, you can solve this problem with bit manipulations. 

```python
def numberOfSteps (self, num: int) -> int:
    steps = 0
    
    while num:
        if num & 1 == 1:
            num ^= 1
        else:
            num >>= 1
            
        steps += 1
        
    return steps

PR #40, opened 2021-04-05
Check in 1528.py

woke7:


1528.py
```diff
@@ -0,0 +1,3 @@
+class Solution:
+    def restoreString(self, s: str, indices: List[int]) -> str:
+        return ''.join([s[indices.index(i)] for i in range(len(indices))])
```
mujingw, line 3: Again, this is an implementation with `O(N^2)` time complexity. Here's why:

1. Finding the index of value `i` in array `indices` (`indices.index(i)`) takes O(N) time
2. You repeated this for N times

On the other hand, the expression `for i in range(len(indices))` has nothing to do with `indices`. You just needed those `N` numbers ranging from `0` to `N - 1` where `N` happens to be the length of the `indices` array (but it is the length of the `s` string too). What I am saying is this expression creates confusion, especially when it is embedded in list comprehension. 

Hint: this problem can be solved using an approach with linear time complexity.

mujingw (approved):
LGTM. A more pythonic way would be:

```python
for char, pos in zip(s, indices):
    res[pos] = char

PR #41, opened 2021-04-06
Check in 1302.py.

woke7:


1302.py
```diff
@@ -0,0 +1,28 @@
...
+                node = queue.popleft()
+                level.append(node.val)
+
+                if node.left:
+                    queue.append(node.left)
+
+                if node.right:
+                    queue.append(node.right)
+
+        return sum(level)
```
mujingw, line 28: I'm concerned about the scope of the `level` variable. It works in Python (and certainly not in Java) but I don't think it's a good practice. The usage of `level` should be limited within the `while` loop. 

Similarly, the following works but I don't think it's a good practice:
```python
for i in range(10):
  # do something here
  # ...
  # ...

print (i) # prints 9

mujingw (approved):
LGTM

PR #42, opened 2021-04-06
Check in 116.py.

woke7:


116.py
```diff
@@ -0,0 +1,35 @@
...
+
+        queue = collections.deque([root])
+
+        while queue:
+            level_length = len(queue)
+
+            for i in range(level_length):
+                node = queue.popleft()
+
+                if i == level_length - 1:
```
mujingw, line 24: This `if-else` is a little awkward, and I am not a fan of accessing `queue[0]` in the loop body. 

Each level is essentially a linked list in this problem. Imagine if you already know your previous node when you are visiting the current node, how would you connect these two nodes together to make them a linked list? Then answer should be trivial. Now the problem becomes how to keep track of your previous node when traversing each level.

116.py
```diff
@@ -0,0 +1,36 @@
...
+        queue = collections.deque([root])
+
+        while queue:
+            level_length = len(queue)
+            prev = None
+
+            for i in range(level_length):
+                node = queue.popleft()
+
+                if prev:
```
mujingw, line 25: Looks great. Concise and correct.

PR #43, opened 2021-04-06
Check in 200.py.

woke7:


200.py
```diff
@@ -0,0 +1,23 @@
...
+
+        return count
+
+    def dfs(self, grid: List[List[str]], row: int, col: int) -> None:
+        if row < 0 or row >= len(grid) or col < 0 or col >= len(grid[0]) or grid[row][col] != '1':
+            return
+
+        grid[row][col] = '0'
+
+        self.dfs(grid, row + 1, col)
```
mujingw, line 20: It's not elegant to hard-code all four calls. Try something like this:

```python
DIR = ((0, 1), (0, -1), (-1, 0), (1, 0))

for dr, dc in DIR:
  self.dfs(grid, row + dr, col + dc)
```
`dr` and `dc` means the delta of between your next position and your current position.

200.py
```diff
@@ -0,0 +1,23 @@
...
+            for j in range(len(grid[0])):
+                if grid[i][j] == '1':
+                    count += 1
+
+                self.dfs(grid, i, j)
+
+        return count
+
+    def dfs(self, grid: List[List[str]], row: int, col: int) -> None:
+        if row < 0 or row >= len(grid) or col < 0 or col >= len(grid[0]) or grid[row][col] != '1':
```
mujingw, line 15: This line is too long. I would break them into two categories:

1. Index out of bound (i.e., position not valid in `grid`)
2. Value does not meet our requirement (i.e., `grid[row][col] != '1'`)

mujingw (approved):
LGTM

PR #45, opened 2021-04-06
Check in 112.py.

woke7:


112.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def maxProfit(self, prices: List[int]) -> int:
+        min_value = prices[0]
+        max_profit = 0
+
+        for i in range(1, len(prices)):
+
```
mujingw, line 7: Redundant whitespace.

woke7: I thought we talked about we want to have space above and below the if-else block?

PR #46, opened 2021-04-06
Check in 695.py.

woke7:


mujingw (commented):
I would leave the exact same comments as I did [here](https://github.com/woke7/leetcode-python/pull/43).

PR #47, opened 2021-04-06
Check in 1254.py.

bunnygod:


mujingw (commented):
This implementation is hard to understand. Here's what I would do:

1. Start with cells on all four borders. If any of these cells is a land cell `0`, mark it as something else (e.g., mark this cell as `2` to differentiate itself from water `1` or a normal land `0`). Use either BFS or DFS to mark all of its adjacent cells too.
2. Go through each cells that are **not** on the border. This time, whatever island you encounter will be a legitimate `closed island` because it is guaranteed that this cell is not adjacent to a border (islands adjacent to borders do not qualify as `closed_island`). This is guaranteed because of what was done in 1: all land cells adjacent to the border have already been marked as `2`. 

Note that once 1 is done, this problem becomes LC-200 Number of Islands.

mujingw:
It looks like this problem is almost identical to [130. Surrounded Regions](https://leetcode.com/problems/surrounded-regions/).

PR #48, opened 2021-04-06
Check in 53.py.

woke7:


53.py
```diff
@@ -0,0 +1,10 @@
+class Solution:
+    def maxSubArray(self, nums: List[int]) -> int:
+        current_max = nums[0]
+        global_max = current_max
+
+        for num in nums[1:]:
+            current_max = max(num, current_max + num)
```
mujingw, line 7: Very smart. It saves one `if-else` block.

mujingw (approved):

PR #49, opened 2021-04-08
Check in 125.py.

woke7:


125.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+class Solution:
+    def isPalindrome(self, s: str) -> bool:
+        regex = re.compile('[^a-zA-Z0-9]')
+        letter_str = regex.sub('', s).lower()
```
mujingw, line 6: No need to use regular expression. This line can be replaced with list comprehension. 

```python
letters = [char.lower() for char in s if char.isalnum()]

File: 125.py [hunk 2]
```diff
@@ -0,0 +1,17 @@
...
+
+        start, end = 0, len(letter_str) - 1
+
+        while start < end:
+            if letter_str[start] != letter_str[end]:
```
mujingw, line 11: This if statement can be merged into `while [condition]:` above.

woke7: Why? this condition is to fail early. If I put it into `while` condition, I will quit `while` and `return True`.

mujingw: Merging internal `if` statements to the `while` statement also allows early exit. Once you exit the while loop, you can check the relationship between `start` and `end` to determine whether it's a palindrome or not. This technique is going to help you in problems like [5. Longest Palindromic Substring](https://leetcode.com/problems/longest-palindromic-substring/).

```python 
def isPalindrome(self, s: str) -> bool:
    letters = [ch.lower() for ch in s if ch.isalnum()]       
    l, r = 0, len(letters) - 1
    
    while l < r and letters[l] == letters[r]:
        l += 1
        r -= 1
    
    return l >= r

woke7: Got you. I did a new implementation without pre-processing the original list.

mujingw: I see. Yes, a one-pass solution is probably better.

PR #51, opened 2021-04-08
Check in 205.py.

woke7:


mujingw (commented):
Way too many repetitiveness. We should extract the common logic into a function and re-use this function when needed.

woke7:
> Way too many repetitiveness. We should extract the common logic into a function and re-use this function when needed.

```python
class Solution:
    def isIsomorphic(self, s: str, t: str) -> bool:
        if len(s) != len(t):
            return False

        dict1, dict2 = dict(), dict()

        if not self.buildMapping(s, t, dict1):
            return False

        if not self.buildMapping(t, s, dict2):
            return False

        if not self.validateMapping(dict1, dict2):
            return False

        if not self.validateMapping(dict2, dict1):
            return False

        return True

    def buildMapping(self, str1: str, str2: str, mapping: dict) -> bool:
        for a, b in zip(str1, str2):
            if a in mapping:
                if mapping[a] != b:
                    return False
            else:
                mapping[b] = a

        return True

    def validateMapping(self, mapping1: dict, mapping2: dict) -> bool:
        for key, value in mapping1.items():
            if mapping2[value] != key:
                return False

        return True

```

I coded something like this, but the result is not correct, could you see a difference? Will dict() be populated in my helper functions?

mujingw:
You have a typo in the `else` block:

```python      
if a in mapping:
    if mapping[a] != b:
        return False
    else:
        mapping[b] = a
```
It should have been:
```python
    else:
        mapping[a] = b
```

I guess this was a copy-paste error.

mujingw:
Also, you don't really need the `validateMapping` function. Removing it won't affect the correctness of your code. And I was expecting something more concise like this:

```python
def isIsomorphic(self, s: str, t: str) -> bool:
    def helper(s, t):
        d = {}
        
        for a, b in zip(s, t):
            if a not in d:
                d[a] = b
            else:
                if d[a] != b:
                    return False
                
        return True
    
    return helper(s, t) and helper(t, s)

PR #53, opened 2021-04-08
Check in 965.py.

woke7:


965.py
```diff
@@ -0,0 +1,27 @@
...
+#     def __init__(self, val=0, left=None, right=None):
+#         self.val = val
+#         self.left = left
+#         self.right = right
+class Solution:
+    def isUnivalTree(self, root: TreeNode) -> bool:
+        if not root:
+            return True
+
+        checker = set([root.val])
```
mujingw, line 12: This set is not needed. You can just compare each child's value with its parent's value whenever you see a child node.

woke7: Yes, I guess I was sleepy.

mujingw (commented):
LGTM

PR #55, opened 2021-04-09
Check in 26.py.

woke7:


26.py
```diff
@@ -0,0 +1,10 @@
+class Solution:
+    def removeDuplicates(self, nums: List[int]) -> int:
+        distinct = 0
+
+        for iterator in range(len(nums)):
+            if nums[distinct] != nums[iterator]:
```
mujingw, line 6: Line 6 is very smart. Here's my version:

```python
def removeDuplicates(self, nums: List[int]) -> int:
    N, slow = len(nums), 0
    
    for fast in range(N):
        if fast > 0 and nums[fast] == nums[fast - 1]:
            continue
        else:
            nums[slow] = nums[fast]
            slow += 1
            
    return slow

PR #56, opened 2021-04-09
Check in 58.py.

woke7:


58.py
```diff
@@ -0,0 +1,15 @@
+class Solution:
+    def lengthOfLastWord(self, s: str) -> int:
+        word_starts = False
+        word_length = 0
+
+        for char in reversed(s):
```
mujingw, line 6: Instead of reversing `s`, why can't we just go through each character from the end of the string all the way back to its beginning?

PR #57, opened 2021-04-09
Check in 617.py.

woke7:


617.py
```diff
@@ -0,0 +1,24 @@
+# Definition for a binary tree node.
+# class TreeNode:
+#     def __init__(self, val=0, left=None, right=None):
+#         self.val = val
+#         self.left = left
+#         self.right = right
+class Solution:
+    def mergeTrees(self, root1: TreeNode, root2: TreeNode) -> TreeNode:
+        merged = None
```
mujingw, line 9: This is not the best location to declare `merged`. You don't need this variable until line 20, which already handles everything you need.

PR #63, opened 2021-04-11
Check in 145.py.

woke7:


mujingw (commented):
I don't have a deep understanding of post-order traversal, so I can't comment on your implementation. My implementation is based on a "double push" approach.

```python
def postorderTraversal(self, root: TreeNode) -> List[int]:
    if not root:
        return []
    
    s, res = [root, root], []
    
    while s:
        curr = s.pop()
        
        if s and s[-1] == curr:
            if curr.right:
                s.extend([curr.right, curr.right])
            
            if curr.left:
                s.extend([curr.left, curr.left])
        else:
            res.append(curr.val)
            
    return res
```

I push every node onto the stack **twice**. When they are popped, the first popped instance is responsible for pushing its children onto the stack; the second instance is responsible for saving its value to the result list.

woke7:
> I don't have a deep understanding of post-order traversal, so I can't comment on your implementation. My implementation is based on a "double push" approach.
> 
> ```python
> def postorderTraversal(self, root: TreeNode) -> List[int]:
>     if not root:
>         return []
>     
>     s, res = [root, root], []
>     
>     while s:
>         curr = s.pop()
>         
>         if s and s[-1] == curr:
>             if curr.right:
>                 s.extend([curr.right, curr.right])
>             
>             if curr.left:
>                 s.extend([curr.left, curr.left])
>         else:
>             res.append(curr.val)
>             
>     return res
> ```
> 
> I push every node onto the stack **twice**. When they are popped, the first popped instance is responsible for pushing its children onto the stack; the second instance is responsible for saving its value to the result list.

I don't quite understand your approach.

PR #65, opened 2021-04-11
Check in 28.py.

woke7:


mujingw (commented):
Overall this is a good pythonic implementation. However, I noticed a couple of redundancies. 

1. Line 3 and 4 are not needed (`needle` could be empty but it's never `None`, and your `while` loop handles the empty case nicely)
2. Line 6 can be simplified to `while idx <= hay_length - nee_length:`

All you need is this:

```python
def strStr(self, haystack: str, needle: str) -> int:
    idx, M, N = 0, len(haystack), len(needle)
    
    while idx <= M - N:
        if all(h == n for h, n in zip(haystack[idx:], needle)):
            return idx
        
        idx += 1
        
    return -1

PR #66, opened 2021-04-12
Check in 3.py.

woke7:


mujingw (approved):
LGTM. Can you generalize your code into a template and apply it to the other sliding window problems?

woke7:
> LGTM. Can you generalize your code into a template and apply it to the other sliding window problems?

I am not sure, I think I need to practice it with more similar problems to be able to come up with a pattern.

PR #67, opened 2021-04-12
Check in 1773.py.

woke7:


mujingw (commented):
The `for` loop is a very clever design, but the if-else branches can be simplified by using a dictionary:

```python
def countMatches(self, items: List[List[str]], ruleKey: str, ruleValue: str) -> int:
    count = 0
    d = {"type": 0, "color": 1, "name": 2}
    
    for item in items:
        if item[d[ruleKey]] == ruleValue:
            count += 1
            
    return count

PR #68, opened 2021-04-12
Check in 1379.py.

woke7:


1379.py
```diff
@@ -0,0 +1,27 @@
...
+
+            if node_o.left:
+                queue_org.append(node_o.left)
+                queue_clo.append(node_c.left)
+
+            if node_o.right:
+                queue_org.append(node_o.right)
+                queue_clo.append(node_c.right)
+
+        return None
```
mujingw, line 27: This line `return None` is not needed. The constraints say: `target node is a node from the original tree`, which means `target` is guaranteed to be valid and you will always have an answer.

PR #70, opened 2021-04-13
Check in 589.py.

woke7:


589.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+
+class Solution:
+    def preorder(self, root: 'Node') -> List[int]:
+        preorder_nodes = []
+        self.get_preorder(root, preorder_nodes)
+        return preorder_nodes
+
+    def get_preorder(self, node: 'Node', preorder_nodes: List[int]):
+        if node:
```
mujingw, line 16: A bit awkward here. I would reverse the logic:

```python
if not node:
    return 

preorder_nodes.append(node.val)

for child in node.children:
    self.get_preorder(child, preorder_nodes)
```
In fact, your `get_preorder` function will never face a case where `node == None`, except for the case when the `root` is `None` in line 10. Therefore, I would probably just handle the `None` case one level up inside the `preorder` function:

```python
def preorder(self, root: 'Node') -> List[int]:
    if not root:
        return []
            
    res = []
    self.get_preorder(res, root)        

    return res

def get_preorder(self, res, root):
    res.append(root.val)
        
    for child in root.children:
        self.get_preorder(res, child)

woke7: But this is assuming that there is no child node that is NULL. It works with this problem's setup, but no guarantee, right? I feel NULL check before access to the variable is more defensive.

mujingw: I agree.

mujingw (approved):

PR #71, opened 2021-04-13
Check in 1207.py.

woke7:


1207.py
```diff
@@ -0,0 +1,12 @@
+class Solution:
+    def uniqueOccurrences(self, arr: List[int]) -> bool:
+        counter = collections.Counter(arr)
+        unique = set()
+
+        for num in counter:
+            if counter[num] in unique:
```
mujingw, line 7: Not pythonic. I would do this:

```python
for num, count in counter.items():
    if count in unique:
    #...

PR #72, opened 2021-04-14
Check in 1221.py.

woke7:


mujingw (commented):
LGTM. But the space complexity can be reduced to O(1):

```python 
def balancedStringSplit(self, s: str) -> int:
    count, res = 0, 0
    
    for ch in s:
        if ch == 'L':
            count += 1
        else:
            count -= 1
            
        if count == 0:
            res += 1
            
    return res

PR #73, opened 2021-04-14
Check in 1588.py.

bunnygod:


1588.py
```diff
@@ -0,0 +1,10 @@
+class Solution:
+    def sumOddLengthSubarrays(self, arr: List[int]) -> int:
+        odd_sum, start, arr_length = 0, 0, len(arr)
+
+        for i in range(arr_length):
+            for j in range(arr_length):
```
mujingw, line 6: 1. `j` doesn't have to start from 0.
2. This problem is not an easy one. During an interview you will probably be asked to implement an O(N) solution, not an O(N^3) one like this.

woke7: This is a brute force solution. I maybe can generate a list of sub-lists via DFS in O(n) complexity?

mujingw: I need some time to think about this problem.

PR #83, opened 2021-04-17
Check in 1688.py

woke7:


1688.py
```diff
@@ -0,0 +1,15 @@
+class Solution:
+    def numberOfMatches(self, n: int) -> int:
```
mujingw, line 2: LGTM. Alternatively, you can implement this in a recursive way:

```python
def numberOfMatches(self, n: int) -> int:
    def dfs(res, teams):
        if teams <= 1:
            return
        
        if teams % 2 == 0:
            games = teams // 2
            res[0] += games
            dfs(res, teams // 2)
        else:
            games = (teams - 1) // 2
            res[0] += games
            dfs(res, 1 + games)
            
    res = [0]
    dfs(res, n)
    
    return res[0]

PR #84, opened 2021-04-17
Check in 704.py.

woke7:


704.py
```diff
@@ -0,0 +1,15 @@
+class Solution:
+    def search(self, nums: List[int], target: int) -> int:
+        left, right = 0, len(nums) - 1
+
+        while left <= right:
```
mujingw, line 5: LGTM. However, if you decide to always use `while left <= right:` going forward, make sure you know the details of this template inside out, including but not limited to:

1. How to avoid infinite loop
2. How to update `left` and `right` inside the loop when the target is not found
3. Upon exiting the loop, what is the relationship between `left` and `right`

PR #86, opened 2021-04-17
Check in 1827.py.

woke7:


1827.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def minOperations(self, nums: List[int]) -> int:
+        operations_num = 0
+
+        if len(nums) <= 1:
+            return 0
+
+        for i in range(1, len(nums)):
+            if nums[i] <= nums[i - 1]:
+                operations_num += (nums[i - 1] - nums[i] + 1)
```
mujingw, line 10: `(nums[i - 1] - nums[i] + 1)` is used twice. Should have given it a name in line 10 and reuse it in line 11.

PR #90, opened 2021-04-17
Check in 1079.py.

woke7:


1079.py
```diff
@@ -0,0 +1,22 @@
+class Solution:
+    def numTilePossibilities(self, tiles: str) -> int:
+        def backtrack(combination: List[int]):
+            if combination:
+                combinations.append(combination)
+
+            for i, tile in enumerate(sorted_tiles):
+                if visited[i] or \
```
mujingw, line 8: As I mentioned in the DFS template, if the condition of the `if` statement is too long, try to break them into several parts for readability.
```python
if used[i]:
    # tile has already been used
    continue
    
if i > 0 and tiles[i] == tiles[i-1] and not used[i-1]:
    # duplicate tile
    continue

PR #91, opened 2021-04-18
Check in 387.py.

woke7:


387.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def firstUniqChar(self, s: str) -> int:
+        letter_idx = collections.defaultdict(list)
+        first_idx = len(s)
+
+        for i, letter in enumerate(s):
+            letter_idx[letter].append(i)
+
+        for value in letter_idx.values():
```
mujingw, line 9: Go through the string once and count the frequency of each character in the string. Once this is done, you know the frequency of any character given to you. Go through the string again: whenever you encounter a character with a frequency that equals to 1, it is the answer. 

```python

def firstUniqChar(self, s: str) -> int:
    c = collections.Counter(s)
    
    for i, ch in enumerate(s):
        if c[ch] == 1:
            return i
        
    return -1

PR #93, opened 2021-04-18
Check in 1832.py.

woke7:


1832.py
```diff
@@ -0,0 +1,12 @@
+class Solution:
+    def checkIfPangram(self, sentence: str) -> bool:
+        alphebet = [False] * 26
+
+        for letter in sentence:
+            alphebet[ord(letter) - ord('a')] = True
+
+        for letter in alphebet:
```
mujingw, line 8: The second `for` loop in your code can be replaced with:

```python
return all(alphabet)
```

Or, you can use a `dict` or a `Counter`:

```python
def checkIfPangram(self, sentence: str) -> bool:
    c = collections.Counter(sentence)
     
    return len(c) == 26
```

Or, you can use a one-liner:
```python
return len(set(sentence)) == 26

PR #94, opened 2021-04-18
Check in 1833.py

woke7:


1833.py
```diff
@@ -3,8 +3,8 @@ def maxIceCream(self, costs: List[int], coins: int) -> int:
         sorted_costs = sorted(costs)
         count = 0
 
-        for i, coin in enumerate(sorted_costs):
-            coins -= coin
+        for cost in sorted_costs:
```
mujingw, line 6: Since `sorted_costs` is only used once in the entire function, you can merge it into the `for` statement and remove line 3:

```python 

for cost in sorted(costs):
```

PR #98, opened 2021-04-19
Check in 238.py.

woke7:


mujingw (approved):
LGTM. Both time complexity and space complexity are optimal.

238.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def productExceptSelf(self, nums: List[int]) -> List[int]:
+        res, left, right = [1] * len(nums), nums[0], nums[len(nums) - 1]
```
mujingw, line 3: I would write it this way:

```python
N = len(nums)
res, left, right = [1] * N, nums[0], nums[-1]

PR #99, opened 2021-04-19
Check in 284.py.

bunnygod:


mujingw (commented):
There are two issues with this implementation:

1. This is an "eager" implementation where you fetch the values from the input data structure `nums` all at once. The problem with this approach is that it has a space complexity of O(N), in addition to the space occupied by the input. 
2. During an interview, the interviewer probably wants to see a "lazy" implementation where you only fetch the next value when you need to. In a "lazy" implementation, the space complexity is O(1), which is better than the "eager" approach.

PR #101, opened 2021-04-19
Check in 155.py.

bunnygod:


mujingw (commented):
This is not an optimal solution. In Python, you can easily save a pair of values into a stack. The second stack in your implementation is not needed.

woke7:
> This is not an optimal solution. In Python, you can easily save a pair of values into a stack. The second stack in your implementation is not needed.

You mean we can push a tuple into one stack, then push a (curr_num, curr_min_num) along the way for every num, right? I felt two stacks make it more straightforward, but sure I'll try and practice it.

mujingw:
> > This is not an optimal solution. In Python, you can easily save a pair of values into a stack. The second stack in your implementation is not needed.
> 
> You mean we can push a tuple into one stack, then push a (curr_num, curr_min_num) along the way for every num, right? I felt two stacks make it more straightforward, but sure I'll try and practice it.

Correct.

PR #103, opened 2021-04-21
Check in 852.py.

woke7:


mujingw (approved):
LGTM

mujingw:
This is actually a hard problem. Normally we compare `arr[start]` (or `arr[end]`) with `arr[mid]`. Here, we have to compare `arr[mid]` with `arr[mid + 1]`, which would not be easy to come up with, even if someone gave us the hint to use binary search. 

There are a few questions we might want to ask ourselves if we come up with this solution without external help. For example, why do we want to compare `arr[mid]` with `arr[mid + 1]`? Why don't we compare `arr[mid - 1]` with `arr[mid]` instead?

bunnygod:
Why? I cannot think of an answer for your question >.<

mujingw:
The second question is slightly easier to answer. `mid + 1` can never be out of bounds because the input array is guaranteed to be a mountain array (i.e., the peak index, which you are searching for, must exist and must **not** be at the right most position of the array). 

Assuming the length of the input array is N, the "mountain array" constraint guarantees that the right-most possible value of `start` is `N - 2`, at which point the loop exits and `start` becomes your final answer [1] (assuming you write your `while` loop and `if` statement in the same way you currently do). This also implies that `start` can only be `N - 3` when it is inside the loop, which further implies that the maximum value of `mid` can only be `N - 2` (why? because the maximum value of `end` can only be `N - 1` by your definition and `mid = (N - 3 + N - 1) // 2` when both `start` and `end` are at their maximum value), which ensures the maximum value of `mid + 1` does not exceed `N - 1` (inclusive), which is always a valid index.

The key takeaway is: the way you choose your **initial** left and right bounds, the way you organize your loop, and the way you shrink your search space all work together. Every time you update any of them, you have to make sure the initial assumption (and definition) before you enter the loop remains true.

Note:
[1] Another way to think of it is: the mountain array's peak index can be `N - 2` at most, and `start` is **always** a possible final answer (because `start` is what your function eventually returns), so `start` can only be `N - 2` (the most extreme case) on exit.

mujingw:
To answer the first question, again, we need to strictly follow the invariants of the `while` loop and the definition of `start` and `end`. 

According to the invariants of the `while` loop, we always have 1) `start < end` inside the loop, and 2) our search space is `[start, end]`, with both end points being inclusive (namely, elements outside of this space are not of our interests). This means, inside the loop, the closest distance `start` and `end` can get is `start == index_i` and `end == index_i + 1`, otherwise (if `start == end`) we would have already exited the loop and returned `start`. This is essentially `start + 1 == end`. Under this circumstance, the new `mid` will be equal to `start` (think about (1 + 2) // 2 == 1, (100 + 101) // 2 == 100, etc.). As a result, `mid - 1` won't even be inside the search space of `[start, end]`. 

On the other hand, `mid + 1` will always stay within the search space by the same reasoning outlined above.

PR #104, opened 2021-04-21
Check in 367.py.

bunnygod:


367.py
```diff
@@ -0,0 +1,15 @@
...
+    def isPerfectSquare(self, num: int) -> bool:
+        start, end = 1, num
+
+        while start <= end:
+            mid = (start + end) // 2
+
+            if mid * mid == num:
+                return True
+            elif mid * mid > num:
+                end = mid - 1
```
mujingw, line 11: Here's how to show the correctness of your algorithm:

1. By definition in line 3, your search space is always `[start, end]` (note: this is an inclusive interval at both the left and the right end points).
2. In line 8, you tested the validity of `mid`, and therefore when line 10 or line 12 is reached, `mid` can't be the answer for sure.
3. In line 11 and line 13, you want to shrink your search space according to the current value of `mid`. Due to the definition in `#1`, your **new** search space for the next iteration should still be an inclusive interval, meaning that both your `start` and `end` should still be possible choices of your final answer. Therefore, it is `end = mid - 1` not `end = mid` and it is `start = mid + 1` not `start = mid`. Invalid answers, i.e., `mid`, cannot stay in your new search space anymore.

PR #108, opened 2021-04-22
Check in 1305.py.

bunnygod:


1305.py
```diff
@@ -0,0 +1,35 @@
+# Definition for a binary tree node.
+# class TreeNode:
+#     def __init__(self, val=0, left=None, right=None):
+#         self.val = val
+#         self.left = left
+#         self.right = right
+class Solution:
+    def getAllElements(self, root1: TreeNode, root2: TreeNode) -> List[int]:
+        def pushleft(node: TreeNode, stack: List[TreeNode]):
+            while node:
```
mujingw, line 10: This `pushleft` template was carefully designed to automatically take care of the `node == None` case using the `while` statement.

1305.py
```diff
@@ -0,0 +1,35 @@
...
+        def addToRes(stack: List[TreeNode], res: List[int]):
+            node = stack.pop()
+            res.append(node.val)
+            pushleft(node.right, stack)
+
+        res, stack1, stack2 = [], [], []
+        pushleft(root1, stack1)
+        pushleft(root2, stack2)
+
+        while stack1 or stack2:
```
mujingw, line 23: LGTM. Did you recognize this is the same pattern we use for the "merge" procedure in merge sort and merge two sorted lists?

1305.py
```diff
@@ -0,0 +1,35 @@
...
+#         self.left = left
+#         self.right = right
+class Solution:
+    def getAllElements(self, root1: TreeNode, root2: TreeNode) -> List[int]:
+        def pushleft(node: TreeNode, stack: List[TreeNode]):
+            while node:
+                stack.append(node)
+                node = node.left
+
+        def addToRes(stack: List[TreeNode], res: List[int]):
```
mujingw, line 14: I can see that you are trying to group these common actions into a method to improve modularity. However, the current design of this method mixed two responsibilities into a single method: 1) add value to `res` and 2) perform `pop` and `push` against `stack`. This kind of design is not recommended.

1305.py
bunnygod: No, I don't, I haven't coded that question.

1305.py
bunnygod: So I should not do this?

PR #110, opened 2021-04-22
Check in 225.py.

bunnygod:


mujingw (commented):
This problem is not hard to implement. I guess what they really want to see are:

1. If you know the trade-offs (in terms of time and space complexity) of each implementation 
2. If you are able to clarify the requirements and priorities with the interviewer and then adjust your design accordingly (e.g., is this stack push-heavy or pop-heavy?)

PR #112, opened 2021-04-23
Check in 21.py.

bunnygod:


mujingw (commented):
LGTM. Same pattern as some of the problems you just solved recently.

PR #113, opened 2021-04-24
Check in 540.py.

bunnygod:


540.py
```diff
@@ -0,0 +1,19 @@
+class Solution:
+    def singleNonDuplicate(self, nums: List[int]) -> int:
+        start, end = 0, len(nums) - 1
+
+        while start < end:
+            mid = start + (end - start) // 2
+
+            if nums[mid - 1] != nums[mid] != nums[mid + 1]:
```
mujingw, line 8: This line seems wrong. `mid` could be equal to `start` (when `start` is immediately adjacent to `end`); therefore, `mid - 1` may not even be in your search space `[start, end]`.

bunnygod: Do you mean when the input has a size of 2? I think that is an invalid input case.

```
You are given a sorted array consisting of only integers where every element appears exactly twice, except for one element which appears exactly once. 
```

mujingw: No, what I mean is the interval `[start, end]` keeps shrinking as your while loop makes progress, and eventually `[start, end]` *could* shrink to an interval with a size of 2, at which point `mid - 1` becomes out of bound.

bunnygod: Hmm, it didn't happen, let me dig into it.

mujingw: Also, the input size of this problem could be `1`: 

```
Constraints:

1 <= nums.length <= 10^5
0 <= nums[i] <= 10^5
```

mujingw: The following is what I just wrote based on what I learned from this [problem](https://github.com/bunnygod/leetcode-python/pull/103#issuecomment-823813063):

```python
def singleNonDuplicate(self, nums: List[int]) -> int:
    N = len(nums)
    l, r = 0, N - 1
    
    while l < r:
        mid = (l + r) // 2
        
        if mid % 2 == 0:
            if nums[mid] == nums[mid + 1]:
                l = mid + 1
            else:
                r = mid
        else:
            if nums[mid] == nums[mid + 1]:
                r = mid - 1
            else:
                l = mid + 1
                
    return nums[l]
```

My code is correct because:

1.  `mid + 1` in my code will never go out of bounds thanks to the definition of `r = N - 1` (even when `l = N - 2` and `r = N - 1`, `mid + 1` is still `N - 1`)
2. `l` and `r` are carefully updated inside the loop to reflect the new interval to be searched
3. Interval `[l, r]` is always shrinking no matter what case it is (this avoids causing an infinite loop during execution)

mujingw: [287. Find the Duplicate Number](https://leetcode.com/problems/find-the-duplicate-number/) is another great problem for more practice.

mujingw (approved):
LGTM

PR #114, opened 2021-04-24
Check in 1603.py.

bunnygod:


1603.py
```diff
@@ -0,0 +1,18 @@
+class ParkingSystem:
+
+    def __init__(self, big: int, medium: int, small: int):
+        self.lots = dict()
```
mujingw, line 4: Since the keys are of integer type, this `dict` can be replaced by an array `[0, big, medium, small]`.

PR #118, opened 2021-04-25
Check in 20.py.

bunnygod:


20.py
```diff
@@ -0,0 +1,14 @@
+class Solution:
+    def isValid(self, s: str) -> bool:
+        stack = []
+        parentheses = {')': '(', ']': '[', '}': '{'}
+
+        for char in s:
+            if char in parentheses.values():
```
mujingw, line 7: `dict.values()` gives you a dictionary view object. However, as the docs says [here](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects), `Values views are not treated as set-like since the entries are generally not unique.)`, line 7 in your code has a time complexity of O(N), which fails to achieve what you intended to do by making `parentheses` a `dict`.

PR #120, opened 2021-04-26
Check in 287.py.

bunnygod:


287.py
```diff
@@ -0,0 +1,14 @@
+class Solution:
+    def findDuplicate(self, nums: List[int]) -> int:
+        sorted_nums = sorted(nums)
+        left, right = 0, len(nums) - 1
+
+        while left < right:
+            mid = left + (right - left) // 2
+
+            if mid + 1 > sorted_nums[mid]:
```
mujingw, line 9: I am not quite sure about the meaning of `mid + 1 > sorted_nums[mid]` here. I use `sorted_num[mid] > mid` as the condition, as shown below:

```python
def findDuplicate(self, nums: List[int]) -> int:
    arr = sorted(nums)
    N = len(arr)
    l, r = 0, N
    
    while l < r:
        mid = (l + r) // 2
        
        if arr[mid] > mid:
            l = mid + 1
        else:
            r = mid
            
    return arr[l]
```

For example, in this sorted array:
```
index 0 1 2| 3 4 5
array 1 2 3| 3 3 4
```

1. When an element `arr[i]` is greater than its index `i`, we know the answer will be to the right of the current element.
2. When an element `arr[i]` is equal to or less than its index, the answer must be to the left of the current element (excluding the current element).

mujingw: Note: in my implementation, `l` and `r` always converge at the first repetitive occurrence (not the first instance of a number, but the first instance of duplicates of this number), at which the while loop exits and index `l` points at the final answer.

bunnygod: Why do you do `r = N` instead of `r = N - 1`?

bunnygod:
```
根据我最后补充的那个note，r = N的话，最后跳出循环时l指向重复元素的第一个重复的值（总体的第二个）。

r如果等于N - 1的话，最后跳出循环时l指向的是重复元素的第一个位置（总体的第一个）。

r = N和r = N - 1虽然都对，但意义不同。
```

PR #121, opened 2021-04-26
Check in 268.py.

bunnygod:


mujingw (approved):
This is a great problem. I found a lot of similarities between this one and [LC-287](https://github.com/bunnygod/leetcode-python/pull/120#discussion_r622283320). The following is my solution based on binary search:

```python
def missingNumber(self, nums: List[int]) -> int:
    arr = sorted(nums)
    l, r = 0, len(nums)
    
    while l < r:
        mid = (l + r) // 2
        
        if arr[mid] > mid:
            r = mid
        else:
            l = mid + 1
        
    return l

PR #122, opened 2021-04-26
Check in 1038.py.

bunnygod:


1038.py
```diff
@@ -0,0 +1,28 @@
...
+        stack, nodes, total_sum = [], [], 0
+        pushleft(stack, root)
+
+        while stack:
+            node = stack.pop()
+            total_sum += node.val
+            nodes.append(node)
+            pushleft(stack, node.right)
+
+        for node in nodes:
```
mujingw, line 23: The beauty of this problem is that you can solve it in one-pass (modifying the value of each node, on the fly, while you traverse the tree) by slightly modifying your `pushleft` function.

1038.py
```diff
@@ -9,20 +9,16 @@ def bstToGst(self, root: TreeNode) -> TreeNode:
         def pushleft(stack: List[TreeNode], node: TreeNode):
```
mujingw, line 9: Should have renamed this function to `pushright` as well.

PR #124, opened 2021-04-27
Check in 1022.py.

bunnygod:


mujingw (approved):
LGTM. You can also solve this problem using the backtracking approach.

PR #125, opened 2021-04-27
Check in 1189.py.

bunnygod:


1189.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def maxNumberOfBalloons(self, text: str) -> int:
+        letter_count = Counter(text)
+        max_num = 0
+
+        while 1:
```
mujingw, line 6: I would avoid writing `while 1` in Python. Moreover, `while True` should be avoided in general, too.

mujingw:
This is what I just wrote:

```python
def maxNumberOfBalloons(self, text: str) -> int:
    def ok(c):
        return c['b'] >= 1 and c['a'] >= 1 and c['l'] >= 2 \
            and c['o'] >= 2 and c['n'] >= 1

    c = Counter(text)
    res = 0
    
    while ok(c):
        for ch in "balloon":
            c[ch] -= 1
            
        res += 1
        
    return res

PR #127, opened 2021-04-27
Check in 169.py.

bunnygod:


mujingw (commented):
This is actually a hard problem. Can you implement an O(N) time complexity and O(1) space complexity solution based on the idea described [here](https://www.cs.utexas.edu/~moore/best-ideas/mjrty/example.html)?

bunnygod:
> This is actually a hard problem. Can you implement an O(N) time complexity and O(1) space complexity solution based on the idea described [here](https://www.cs.utexas.edu/~moore/best-ideas/mjrty/example.html)?

Interesting idea, let me try to implement the idea described.

169.py
```diff
@@ -1,4 +1,21 @@
...
+#     def majorityElement(self, nums: List[int]) -> int:
+#         sorted_nums = sorted(nums)
+#         return sorted_nums[len(nums) // 2]
+
+# O(n) time complexity.
 class Solution:
     def majorityElement(self, nums: List[int]) -> int:
-        sorted_nums = sorted(nums)
-        return sorted_nums[len(nums) // 2]
+        count = [nums[0], 1]
```
mujingw, line 9: `count[0]` and `count[1]` are not descriptive. I would rather declare two variables `current_majority` and `current_count` for better readability.

PR #128, opened 2021-04-27
Check in 63.py.

bunnygod:


63.py
```diff
@@ -0,0 +1,23 @@
+class Solution:
+    def uniquePathsWithObstacles(self, obstacleGrid: List[List[int]]) -> int:
+        m, n = len(obstacleGrid), len(obstacleGrid[0])
+        dp = [[0] * n for _ in range(m)]
+        print(dp)
```
mujingw, line 5: This line should be deleted.

63.py
```diff
@@ -0,0 +1,23 @@
...
+        for i in range(m):
```
mujingw, line 6: This `for` loop and `break` structure can be replaced by a `while` loop. In general, having a `break` or `continue` inside a loop decreases readability.

PR #130, opened 2021-04-28
Check in 234.py.

bunnygod:


mujingw (approved):
Other than the lack of a "list recovery" step as we already discussed, this problem can be solved by taking the advantage of this constraint: `0 <= Node.val <= 9`

```python
def isPalindrome(self, head: ListNode) -> bool:
    idx_from_left = 0
    curr = head
    lookup = [0 for _ in range(10)]
    
    while curr:
        lookup[curr.val] += idx_from_left
        idx_from_left += 1
        curr = curr.next
    
    curr = head
    idx_from_right = idx_from_left - 1
    
    while curr:
        lookup[curr.val] -= idx_from_right
        idx_from_right -= 1
        curr = curr.next
        
    return all(count == 0 for count in lookup)

mujingw:
This is also an `O(1)` space solution because `lookup` only takes a finite amount of space (10 slots).

bunnygod:
> This is also an `O(1)` space solution because `lookup` only takes up a finite amount of space (10 slots).

These math tricks are kind of confusing.

PR #131, opened 2021-04-28
Check in 530.py.

bunnygod:


mujingw (approved):
LGTM. As an extra practice, try to rewrite it in a recursive way.

PR #135, opened 2021-04-30
Check in 1451.py.

bunnygod:


mujingw (commented):
Your current implementation involves too much redundant work, most of which can be delegated to built-in functions. The following is what I just wrote, which is faster than 99.82% of all Python3 submissions.

```python
def arrangeWords(self, text: str) -> str:
    words = text.split()
    words[0] = words[0].lower()
    words.sort(key=len)
    words[0] = words[0].capitalize()
    
    return " ".join(words)  
```

The _key_ here is to use `len` as our `key` in the `sort` function. The problem also says `if two words have the same length, arrange them in their original order`, which coincides with the behavior of the built-in `list.sort()` function, so we are taking an advantage of it here and saved some extra handling. I am mentioning this point because I want to emphasize that even though built-in functions come in handy most of the time, we should still know whether or not they are applicable to the problem we are trying to solve (and in this particular LeetCode problem the `list.sort()` by `key=len` utility happens to be applicable). 

Note: My solution from just two months ago was also very lengthy (and slow). I am pasting it here to show you what mistakes I made last time. You can see clear differences (readability, modularity, and efficiency) between my new version and my old version. 

```python 
def arrangeWords(self, text: str) -> str:
    words = sorted([(w.lower(), len(w)) for w in text.split()], key=lambda x: x[1])
    res = []
    
    for i, (word, l) in enumerate(words):
        res.extend(list(word))
        
        if i != len(words) - 1:
            res.extend(" ")
        
    res[0] = res[0].upper()
    
    return "".join(res)
```

1. It tried to `lower()` every word, which was unnecessary. Only the first word needs to be "lowered".
2. It built a sorted list `words`. The length of each word was also saved in this list, which required extra space. This sort by length operation can be easily done by using the `len()`  [built-in function](https://docs.python.org/3/library/functions.html) as the `key` in sorting, and all the hard work will be handled by the built-in `list.sort()` [function](https://docs.python.org/3/library/stdtypes.html#list.sort).
3. The worst part was that it broke up all strings into lists of characters and then joined all these characters back into a string.

bunnygod:
> Your current implementation involves too much redundant work, most of which can be delegated to built-in functions. The following is what I just wrote, which is faster than 99.82% of all Python3 submissions.
> 
> ```python
> def arrangeWords(self, text: str) -> str:
>     words = text.split()
>     words[0] = words[0].lower()
>     words.sort(key=len)
>     words[0] = words[0].capitalize()
>     
>     return " ".join(words)  
> ```
> 
> The _key_ here is to use `len` as our `key` in the `sort` function. The problem also says `if two words have the same length, arrange them in their original order`, which coincides with the behavior of the built-in `list.sort()` function, so we are taking an advantage of it here and saved some extra handling. I am mentioning this point because I want to emphasize that even though built-in functions come in handy most of the time, we should still know whether or not they are applicable to the problem we are trying to solve (and in this particular LeetCode problem the `list.sort()` by `key=len` utility happens to be applicable).
> 
> Note: My solution from just two months ago was also very lengthy (and slow). I am pasting it here to show you what mistakes I made last time. You can see clear differences (readability, modularity, and efficiency) between my new version and my old version.
> 
> ```python
> def arrangeWords(self, text: str) -> str:
>     words = sorted([(w.lower(), len(w)) for w in text.split()], key=lambda x: x[1])
>     res = []
>     
>     for i, (word, l) in enumerate(words):
>         res.extend(list(word))
>         
>         if i != len(words) - 1:
>             res.extend(" ")
>         
>     res[0] = res[0].upper()
>     
>     return "".join(res)
> ```
> 
> 1. It tried to `lower()` every word, which was unnecessary. Only the first word needs to be "lowered".
> 2. It built a sorted list `words`. The length of each word was also saved in this list, which required extra space. This sort by length operation can be easily done by using the `len()`  [built-in function](https://docs.python.org/3/library/functions.html) as the `key` in sorting, and all the hard work will be handled by the built-in `list.sort()` [function](https://docs.python.org/3/library/stdtypes.html#list.sort).
> 3. The worst part was that it broke up all strings into lists of characters and then joined all these characters back into a string.

The key part is ``words.sort(key=len)` won't alter original order of words with same length?

mujingw:
> > Your current implementation involves too much redundant work, most of which can be delegated to built-in functions. The following is what I just wrote, which is faster than 99.82% of all Python3 submissions.
> > ```python
> > def arrangeWords(self, text: str) -> str:
> >     words = text.split()
> >     words[0] = words[0].lower()
> >     words.sort(key=len)
> >     words[0] = words[0].capitalize()
> >     
> >     return " ".join(words)  
> > ```
> > 
> > 
> > The _key_ here is to use `len` as our `key` in the `sort` function. The problem also says `if two words have the same length, arrange them in their original order`, which coincides with the behavior of the built-in `list.sort()` function, so we are taking an advantage of it here and saved some extra handling. I am mentioning this point because I want to emphasize that even though built-in functions come in handy most of the time, we should still know whether or not they are applicable to the problem we are trying to solve (and in this particular LeetCode problem the `list.sort()` by `key=len` utility happens to be applicable).
> > Note: My solution from just two months ago was also very lengthy (and slow). I am pasting it here to show you what mistakes I made last time. You can see clear differences (readability, modularity, and efficiency) between my new version and my old version.
> > ```python
> > def arrangeWords(self, text: str) -> str:
> >     words = sorted([(w.lower(), len(w)) for w in text.split()], key=lambda x: x[1])
> >     res = []
> >     
> >     for i, (word, l) in enumerate(words):
> >         res.extend(list(word))
> >         
> >         if i != len(words) - 1:
> >             res.extend(" ")
> >         
> >     res[0] = res[0].upper()
> >     
> >     return "".join(res)
> > ```
> > 
> > 
> > 
> > 1. It tried to `lower()` every word, which was unnecessary. Only the first word needs to be "lowered".
> > 2. It built a sorted list `words`. The length of each word was also saved in this list, which required extra space. This sort by length operation can be easily done by using the `len()`  [built-in function](https://docs.python.org/3/library/functions.html) as the `key` in sorting, and all the hard work will be handled by the built-in `list.sort()` [function](https://docs.python.org/3/library/stdtypes.html#list.sort).
> > 3. The worst part was that it broke up all strings into lists of characters and then joined all these characters back into a string.
> 
> The key part is `words.sort(key=len)` won't alter original order of words with same length?

That's right. 

1. `Sorts are guaranteed to be stable. That means that when multiple records have the same key, their original order is preserved.`
2. `The sort is stable – if two items have the same key, their order will be preserved in the sorted list.`

References:

1. [Sort Stability and Complex Sorts](https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts)
2. [The Old Way Using Decorate-Sort-Undecorate](https://docs.python.org/3/howto/sorting.html#the-old-way-using-decorate-sort-undecorate)

PR #136, opened 2021-04-30
Check in 1119.py.

bunnygod:


mujingw (commented):
This code demonstrates that you have learned how and when to use the `str.join()` [function](https://docs.python.org/3/library/stdtypes.html#str.join) and list comprehension. 

However, line 3 is still not pythonic. I would write something like this instead, which saves a lot of typing. 
```python
vowels = set("aeiou")
```

Also, I found this problem very similar to [LC-283](https://leetcode.com/problems/move-zeroes/). The difference is that this problem asks to return a new string while LC-283 asks to modify `in-place`.

PR #139, opened 2021-04-30
Check in 896.py.

bunnygod:


mujingw (approved):
LGTM

mujingw:
On a side note, the following paradigm is what we usually do to generate and process every pair of **adjacent** elements in an array in Python. Depending on one's familiarity with Python, it might be _slightly_ more readable. 

```python
def isMonotonic(self, A: List[int]) -> bool:
    asc, desc = True, True
    
    for a, b in zip(A, A[1:]):
        if a < b:
            desc = False
        elif a > b:
            asc = False
            
    return asc or desc
```

However, we should be aware that creating `zip(A, A[1:])` already takes one-pass on the array and it is not as efficient as your current implementation.

PR #140, opened 2021-05-01
Check in 346.py.

bunnygod:


346.py
```diff
@@ -0,0 +1,24 @@
+class MovingAverage:
+
+    def __init__(self, size: int):
+        """
+        Initialize your data structure here.
+        """
+        self.size = size
+        self.queue = collections.deque()
+        self.num_sum = 0
+        self.num_count = 0
```
mujingw, line 10: `self.num_count` is not needed. You can always use `len(self.queue)` to get this information.

PR #141, opened 2021-05-01
Check in 678.py.

bunnygod:


mujingw (approved):
I don't really understand this solution, but the time complexity looks OK.

mujingw:
Sharing my three-pass (go through the string once, reverse the string, and then go through the string again) solution:

```python
def checkValidString(self, s: str) -> bool:
    def check_char(target_ch, s):
        count, comp_count = 0, 0
        
        for ch in s:
            if ch == target_ch or ch == "*":
                count += 1
            else:
                comp_count += 1
        
            if count < comp_count:
                return False
        
        return True
    
    return check_char("(", s) and check_char(")", s[::-1])

PR #142, opened 2021-05-01
Check in 415.py.

bunnygod:


415.py
```diff
@@ -0,0 +1,18 @@
...
+            value = (x1 + x2 + carry) % 10
+            carry = (x1 + x2 + carry) // 10
+            res.append(value)
+            p1 -= 1
+            p2 -= 1
+
+        if carry > 0:
+            res.append(carry)
+
+        return ''.join(str(x) for x in res[::-1])
```
mujingw, line 18: If your `res` is a queue data structure, you can avoid the step of reversing it here.

bunnygod: How to do that? Why does it matter it is a queue or a stack?

mujingw: Instead of appending the digits to the end of the list, you can insert the digits to the front of the list. In this way, you don't need to do `res[::-1]`, which requires going through the `res` list one more time.

Inserting elements to the front of a `list` is expensive, so you would want to use `collections.deque` and the `appendleft()` method. 

```python
def addStrings(self, num1: str, num2: str) -> str:
    p1, p2 = len(num1) - 1, len(num2) - 1
    carry = 0
    res = deque([])
    
    while p1 >= 0 and p2 >= 0:
        d1, d2 = ord(num1[p1]) - ord('0'), ord(num2[p2]) - ord('0')
        res.appendleft(str((d1 + d2 + carry) % 10))
        carry = (d1 + d2 + carry) // 10
        p1 -= 1
        p2 -= 1
        
    while p1 >= 0:
        d = ord(num1[p1]) - ord('0')
        res.appendleft(str((d + carry) % 10))
        carry = (d + carry) // 10
        p1 -= 1
        
    while p2 >= 0:
        d = ord(num2[p2]) - ord('0')
        res.appendleft(str((d + carry) % 10))
        carry = (d + carry) // 10
        p2 -= 1
    
    if carry > 0:
        res.appendleft("1")
    
    return "".join(res)

PR #146, opened 2021-05-03
Check in 917.py.

bunnygod:


mujingw (approved):
LGTM. This should be the "two-pointer" template for solving many other similar string and array problems.

PR #147, opened 2021-05-03
Check in 509.py.

bunnygod:


mujingw (approved):
LGTM. Try to solve this using an iterative approach.

bunnygod:
> LGTM. Try to solve this using an iterative approach.

Yes, I will do it. The recursive code won't get a hire I think.

PR #148, opened 2021-05-03
Check in 706.py.

bunnygod:


706.py
```diff
@@ -0,0 +1,49 @@
...
+        Initialize your data structure here.
+        """
+        self.space_size = 9999
+        self.space = [[] for _ in range(self.space_size)]
+
+    def put(self, key: int, value: int) -> None:
+        """
+        value will always be non-negative.
+        """
+        hashcode = key % self.space_size
```
mujingw, line 14: This line of code is repeated three times in this class. We should extract it into its own function for maintainability (e.g., what if we later decided to use a different hash function for this hash map design?).

706.py
```diff
@@ -0,0 +1,49 @@
...
+
+        for idx, [k, v] in enumerate(self.space[hashcode]):
```
mujingw, line 16: A key-value pair should be something immutable (namely, once a pair is stored in the hash map, the content of this pair should not be allowed to be modified in any way). Using a `tuple` (i.e., `(k, v)`) instead of a `list` makes more sense here.

PR #149, opened 2021-05-04
Check in 705.py.

bunnygod:


mujingw (approved):
LGTM

705.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Initialize your data structure here.
+        """
+        self.SPACE_SIZE = 9999
+        self.space = [[] for _ in range(self.SPACE_SIZE)]
+
+    def add(self, key: int) -> None:
+        hashcode = self.computeHash(key)
+
+        for k in self.space[hashcode]:
```
mujingw, line 17: This `for k in self.space[loc]:` operation has a linear time complexity. It takes `O(N)` time to find `k` from `self.space[loc]`, assuming there are `N` elements in your linked list `self.space[loc]`. If somehow all the elements collide into a specific slot in `self.space`, the overall performance of your current hash set design degrades to a simple linked list's performance.

This same issue exists in your `remove()` and `contains()` implementations as well.

705.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.SPACE_SIZE = 9999
+        self.space = [[] for _ in range(self.SPACE_SIZE)]
+
+    def add(self, key: int) -> None:
+        hashcode = self.computeHash(key)
```
mujingw, line 15: Related to what I just commented [here](https://github.com/bunnygod/leetcode-python/pull/149#discussion_r660698239), this variable does not mean or represent a **_hash_** or **_hash code_**; it means a location within the `self.space` array. 

A _**hash**_ is a value calculated based on one or more properties of a `key` object, and it can be obtained by calling `hash(a_key_object)` in Python. We then use this hash value to derive a location in the `self.space` array to store the `key` object; this location is what we need here to store the `key`. Therefore, there are two steps involved:

1. Compute the hash code of the `key` object
2. Map the hash code to a location in the `self.space` array

I want to emphasize the difference between a hash and a location in the array of buckets: we rely on the location index to identify the correct slot in the array of buckets to store the `key` object; we do not use an object's hash code to identify this slot. 

Currently, you are mixing these two concepts together. Your current implementation works only because your key object happens to be an `int` object whose value _can_ be used as a hash code. Imagine if your hash set implementation has to deal with objects of the `Animal` type as the keys, can you still use `dog % self.SPACE_SIZE` to get your location index? No, you will have to get the hash code of `dog` using `hash(dog)` first, and then map this hash code to a location index. What you return to `add()`, `remove()`, and `contains()` should always be a location index, not a hash code.

705.py
```diff
@@ -0,0 +1,50 @@
+# N.B.
+# We should go back visit linkedlist and BST solutions.
```
mujingw, line 2: Great! Looks like you were already aware of what we need to do to fully digest this design hash set problem.

705.py
```diff
@@ -0,0 +1,50 @@
...
+        hashcode = self.computeHash(key)
+
+        for k in self.space[hashcode]:
+            if key == k:
+                return
+
+        self.space[hashcode].append(key)
+
+    def remove(self, key: int) -> None:
+        hashcode = self.computeHash(key)
```
mujingw, line 24: Looking at the source code of `HashMap.java` in Java 8, we can tell that hash, hash code, and bucket location are three different concepts. [Line 336](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/HashMap.java#l336) is where the `hash` is computed based on the `key` object's `hashCode()`. [Line 569](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/HashMap.java#l569) shows how the `hash` is mapped to a location in the bucket, which is the `tab` variable in the source code. `tab` is a local alias of the `table` variable declared in [line 395](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/HashMap.java#l395).

PR #150, opened 2021-05-04
Check in 905.py.

bunnygod:


905.py
```diff
@@ -0,0 +1,15 @@
+class Solution:
+    def sortArrayByParity(self, A: List[int]) -> List[int]:
+        i, j = 0, len(A) - 1
+
+        while i < j:
+            if A[i] % 2 > A[j] % 2:
```
mujingw, line 6: I can tell that this line is trying to determine whether `A[i]` is odd and `A[j]` is even because `1 > 0`. However, this meaning is not explicitly conveyed through the code itself--it requires an extra step for the reader to figure it out.

905.py
```diff
@@ -0,0 +1,15 @@
...
+                A[i], A[j] = A[j], A[i]
+
+            if A[i] % 2 == 0:
```
mujingw, line 9: It seems that you are more comfortable incrementing indexes one position at a time. I like to increment them all at once as long as the loop invariant still holds true.

```python
while l < r and A[l] % 2 == 0:
    l += 1
    
while l < r and A[r] % 2 != 0:
    r -= 1
```

There is no difference in terms of time complexity or performance; it's more of a coding style issue.

PR #152, opened 2021-05-05
Check in 1705.py.

bunnygod:


mujingw (commented):
This is problem [1704](https://leetcode.com/problems/determine-if-string-halves-are-alike/), not 1705.

mujingw (approved):
LGTM

PR #153, opened 2021-05-06
Check in 11.py.

bunnygod:


mujingw (approved):
LGTM. This problem is easy to code but hard to prove.

mujingw:
Your two pointer approach is hard to prove. A binary search approach like the one below would be easier to explain during an interview. 

```python
def maxArea(self, height: List[int]) -> int:
    def scan_from_left(height):
        res = 0
        s = []
        h_to_idx = defaultdict(list)

        for i, h in enumerate(height):
            if not s or h > s[-1]:
                s.append(h)            
                h_to_idx[h].append(i)
            else:
                loc_in_s = bisect.bisect_left(s, h) # bar slightly taller than but farthest from current bar
                width = i - h_to_idx[s[loc_in_s]][0]
                res = max(res, width * h)
        
        return res
    
    return max(scan_from_left(height), scan_from_left(height[::-1]))

PR #154, opened 2021-05-06
Check in 122.py.

bunnygod:


122.py
```diff
@@ -0,0 +1,9 @@
+class Solution:
+    def maxProfit(self, prices: List[int]) -> int:
+        max_profit = 0
```
mujingw, line 3: There is no comparison involved throughout this function body. I don't think it's appropriate to name it `max_profit` (is it a maximum of what?). It's just the total profit you can make by deploying some strategy.

PR #155, opened 2021-05-06
Check in 162.py.

bunnygod:


mujingw (approved):
LGTM. But make sure you know how to prove (or show) that your code is correct by walking the interviewer through your code.

bunnygod:
> LGTM. But make sure you know how to prove (or show) that your code is correct by walking the interviewer through your code.

What is a concise, clear, accurate way to explain why my binary search is correct? Could you show me an example?

mujingw:
You should always ask yourself these questions and show the interviewers you have answered these questions in your code:

1. Is the initial state correct and appropriate to the problem?
2. Is the search interval shrinking in every loop?
3. After shrinking the search interval, does the initial assumption still hold? To make sure the update/shrinking is correctly done, you have to make sure whatever assumption you make initially still holds true after the update/shrinking.
4. What does it mean when the loop exits? 

When I did my code reviews, I asked these questions in my head as well. 

For concrete examples, you can refer to [here](https://github.com/bunnygod/leetcode-python/pull/103#issuecomment-825269172) and [here](https://github.com/bunnygod/leetcode-python/pull/103#issuecomment-825295499).

mujingw:
I just noticed that this [post](https://leetcode.com/discuss/general-discussion/786126/Python-Powerful-Ultimate-Binary-Search-Template.-Solved-many-problems) happened to ask the same questions as I did:

```
Some of the most common problems include:

- When to exit the loop? Should we use left < right or left <= right as the while loop condition?
- How to initialize the boundary variable left and right?
- How to update the boundary? How to choose the appropriate combination from left = mid, left = mid + 1 and right = mid, right = mid - 1?
```

PR #156, opened 2021-05-07
Check in 252.py.

bunnygod:


mujingw (approved):
This implementation looks OK. However, I would suggest that you invest some time on learning the following template, which would allow you to solve many other interval-related problems with minimum effort. 

```python
def canAttendMeetings(self, intervals: List[List[int]]) -> bool:
    d = defaultdict(int)
    
    for start, end in intervals:
        d[start] += 1
        d[end] -= 1
        
    total = 0
    
    for time in sorted(d.keys()):
        total += d[time]
        
        if total > 1:
            return False
        
    return True

mujingw:
On a separate note, `zip(A, A[1:])` is more pythonic but I am concerned about its performance. On the surface it seems that it involves creating a new list `A[1:]` through list slicing, but I'm not sure if there's optimization behind the scene. For now I'd say: if two lists already exist in parallel (same length, and you want to visit elements at the same index simultaneously), using `zip` would be a good idea; if we have to *manually* create the second list in order to use `zip`, we might want to think twice.

mujingw:
> > This implementation looks OK. However, I would suggest that you invest some time on learning the following template, which would allow you to solve many other interval-related problems with minimum effort.
> > ```python
> > def canAttendMeetings(self, intervals: List[List[int]]) -> bool:
> >     d = defaultdict(int)
> >     
> >     for start, end in intervals:
> >         d[start] += 1
> >         d[end] -= 1
> >         
> >     total = 0
> >     
> >     for time in sorted(d.keys()):
> >         total += d[time]
> >         
> >         if total > 1:
> >             return False
> >         
> >     return True
> > ```
> 
> This is the so-called sweep-line methodology? I cannot make full sense of it. Could you help me understand it?

You just need to draw a diagram of X-axis to visualize what's going on. Every time you encounter a `start` of an interval, you need to increment your running total (because at this point there is one more meeting that is going to start; no matter what's going on elsewhere, you know for sure one more meeting is going to start right now); every time you encounter an `end` of an interval, you need to decrement your running total (because at this point you know one meeting just ended). Running total in this problem means the number of meetings that are ongoing at this point in time.

Obviously running total starts at 0 because there's no meeting going on at the very beginning. And then you walk through the X-axis from left to right to see if there is anything (any event) triggering you to increment or decrement this running total. Each key in the dictionary is an event, and each event could be a "start a new meeting" event or a "finishing an existing meeting" event. You can tell which event it is by looking at the value of this key: `+1` means a starting/incrementing trigger; `-1` means a finishing/decrementing trigger. You adjust your running total based on these triggers accordingly.

For this particular problem, the running total should never exceed 1, otherwise it means that an overlap of meetings occurs, which makes the person unable to attend all meetings.

mujingw (commented):
LGTM

PR #157, opened 2021-05-07
Check in 1389.py.

bunnygod:


mujingw (commented):
This problem presents a perfect case to use `zip`:

```python
def createTargetArray(self, nums: List[int], index: List[int]) -> List[int]:
    res = []
    
    for num, idx in zip(nums, index):
        res.insert(idx, num)
        
    return res

PR #158, opened 2021-05-07
Check in 897.py.

bunnygod:


mujingw (approved):
LGTM. Your `res` acts as a dummy node as seen in many linked list problems.

This problem should be marked as "Medium`.

PR #159, opened 2021-05-07
389

bunnygod:


mujingw (approved):
LGTM

mujingw:
This is a map-reduce problem in functional programming. We `map` all the characters in `s` and `t` to their ASCII values and store them into a list, and then `reduce` the elements in this list by applying `^` to them.  

```python
def findTheDifference(self, s: str, t: str) -> str:
    def single_number(nums):
        return reduce(lambda x, y: x ^ y, nums)
    
    return chr(single_number([ord(ch) for ch in s + t]))

bunnygod:
> ```python
> reduce
> ```

This is nuts, how did you learn it?

mujingw:
> > ```python
> > reduce
> > ```
> 
> This is nuts, how did you learn it?

I don't know much about it. All I know is that it's a cool way to get the product or sum of _all elements_ in an array. In this particular problem, `^` has properties similar to those of addition and multiplication.

PR #160, opened 2021-05-07
Check in 151.py.

bunnygod:


mujingw (approved):
LGTM. One-liner just for fun:

```python
def reverseWords(self, s: str) -> str:
    return " ".join(s.lstrip().rstrip().split()[::-1])
```

On a more serious note, this problem is much more difficult if we have to handle the dirty work of strip, split, and swap by ourselves. The idea is simple:

1. Remove all whitespaces from both ends.
2. Reverse the entire string.
3. Locate and reverse each group of characters between whitespaces and make sure all the extra whitespaces are also removed.

The implementation of this idea is tricky, especially in step 3.

bunnygod:
> LGTM. One-liner just for fun:
> 
> ```python
> def reverseWords(self, s: str) -> str:
>     return " ".join(s.lstrip().rstrip().split()[::-1])
> ```
> 
> On a more serious note, this problem is much more difficult if we have to handle the dirty work of strip, split, and swap by ourselves. The idea is simple:
> 
> 1. Remove all whitespaces from both ends.
> 2. Reverse the entire string.
> 3. Locate and reverse each group of characters between whitespaces and make sure all the extra whitespaces are also removed.
> 
> The implementation of this idea is tricky, especially in step 3.

Will we be expected to not use any build-in function to do this problem in a real interview?

mujingw:
> > LGTM. One-liner just for fun:
> > ```python
> > def reverseWords(self, s: str) -> str:
> >     return " ".join(s.lstrip().rstrip().split()[::-1])
> > ```
> > 
> > 
> > On a more serious note, this problem is much more difficult if we have to handle the dirty work of strip, split, and swap by ourselves. The idea is simple:
> > 
> > 1. Remove all whitespaces from both ends.
> > 2. Reverse the entire string.
> > 3. Locate and reverse each group of characters between whitespaces and make sure all the extra whitespaces are also removed.
> > 
> > The implementation of this idea is tricky, especially in step 3.
> 
> Will we be expected to not use any build-in function to do this problem in a real interview?

It depends. If the role you interview for is more toward the system level, then you are expected to solve this problem using only pointers and/or indexes. If you are interviewing for a full-stack development role, then you are expected to show that you are familiar with all the tools the standard library provides.

PR #161, opened 2021-05-08
Check in 186.py.

bunnygod:


186.py
```diff
@@ -0,0 +1,27 @@
...
+            while start < end:
+                s[start], s[end] = s[end], s[start]
+                start += 1
+                end -= 1
+
+        N = len(s)
+        reverse(0, N - 1)
+        start, end = 0, 0
+
+        while end < N:
```
mujingw, line 16: The logic inside your `while` loop is a bit complicated. All it needs to do are: 

1. For every `start`, find the next whitespace or the end of the string.
2. Reverse all the elements between the current `start` and the next whitespace.
3. Set `start` to the next word.
4. Repeat.

```python
def reverseWords(self, s: List[str]) -> None:
    """
    Do not return anything, modify s in-place instead.
    """
    def reverse(arr, l, r):
        while l < r:
            arr[l], arr[r] = arr[r], arr[l]
            l += 1
            r -= 1
            
    N = len(s)
    reverse(s, 0, N - 1)
    p = 0
    
    while p < N:
        q = p
        
        while q < N and s[q].isalpha():
            q += 1
            
        reverse(s, p, q - 1)
        
        p = q + 1

File: 186.py [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+            if s[end] == ' ':
```
mujingw, line 17: Line 17 to line 25 can be simplified to:

```python
if s[end] == ' ' or end == N - 1:
    reverse(start, end - 1)
    start = end

end += 1

File: 186.py [hunk 3]
```diff
@@ -16,12 +16,10 @@ def reverse(start: int, end: int) -> None:
         while end < N:
             if s[end] == ' ':
                 reverse(start, end - 1)
-                end += 1
-                start = end
+                start = end + 1
             elif end == N - 1:
```
mujingw, line 20: The case of `end == N -  1` is not that special. It can be merged into the `s[end] == ' '` case.

mujingw: Never mind. I misread your logic. This implementation is good.

PR #163, opened 2021-05-08
Check in 543.py.

bunnygod:


543.py
```diff
@@ -0,0 +1,31 @@
...
+#         self.val = val
+#         self.left = left
+#         self.right = right
+class Solution:
+    def diameterOfBinaryTree(self, root: TreeNode) -> int:
+        def getMaxDepth(node: TreeNode) -> int:
+            if not node:
+                return 0
+
+            if not node.left and not node.right:
```
mujingw, line 13: Line 13 is not needed to make this `getMaxDepth` function correct. If both children are `None`, line 16 (as well as the subsequent recursive calls from line 16) handles it correctly.

Line 10 is needed because a `None` node will cause `node.left` and `node.right` to be incorrect.

You may argue that line 13 enables early return and saves two more calls to the next level. My counter argument is that we need to separate concerns between the two levels of recursive calls: a `None` node should take care of itself (`if not node: return 0`); its parent should not interfere at the parent level. 

Nowadays, the benefit we get from improving readability and maintainability and thus saving developers' time outweigh the benefit we get from saving two functions calls for the machine. I want to make sure whatever template you use, it is the simplest and correct template. Simpler means easier to memorize and less chance to introduce error.

543.py
```diff
@@ -0,0 +1,31 @@
...
+                return 1
+
+            return max(getMaxDepth(node.left), getMaxDepth(node.right)) + 1
+
+        stack = [root]
+        diameter = 0
+
+        while stack:
```
mujingw, line 21: Did you notice this structure (from line 18 to line 29) is very similar to the BFS template?

bunnygod: This is pre-order traversal, right?

mujingw: It looks like it's post-order--you get your results from both children first and then update `diameter` of the current node.

mujingw: This `while stack:` structure is an iterative DFS template.

PR #164, opened 2021-05-08
Check in 1064.py.

bunnygod:


mujingw (approved):
This is a very good problem for practicing binary search. The template you use in this problem is my favorite and I can confirm that line 9, 11, and 13 are all consistent with line 3 and they are all mathematically correct.

Line 15 is also crucial in this template because you can't determine whether or not you have located the final answer until you exit the `while` loop. The **binary search** algorithm here only helps you narrow your search interval down to size one, which then makes it _trivial_ to get to the final answer; it is not meant to help you directly locate the final answer in the middle of the loop execution. Without binary search, your search interval would be the entire array and you would have to make many more comparisons (at the `O(N)` magnitude) to get to the final answer (with binary search, you make `O(log N) + 1` comparisons. 

Note: You cannot be sure that you have found the final answer while you are still inside the loop (even when `arr[mid] == mid`, you still don't know whether you have found the *smallest*, which is what this problem asks).

PR #165, opened 2021-05-08
Check in 34.py.

bunnygod:


34.py
```diff
@@ -0,0 +1,31 @@
+class Solution:
+    def findIndex(self, nums: List[int], target: int, is_first: bool) -> int:
```
mujingw, line 2: `is_first` is not descriptive. This variable indicates whether the caller wants to search for the left most or the right most instance of `target` in the `nums` array.

34.py
```diff
@@ -0,0 +1,31 @@
...
+        N = len(nums)
+        left, right = 0, N - 1
+
+        while left <= right:
```
mujingw, line 6: I'm worried because you use yet another template this time. I don't think you would be able to recall or identify which is which during a contest or an interview. 

In line 6, why do you use `<=` this time?

34.py
```diff
@@ -0,0 +1,31 @@
...
+                        left = mid + 1
+            elif nums[mid] < target:
+                left = mid + 1
+            else:
+                right = mid - 1
+
+        return -1
+
+    def searchRange(self, nums: List[int], target: int) -> List[int]:
+        first_idx = self.findIndex(nums, target, True)
```
mujingw, line 28: This part is a good design because you unify two functionalities (search left most and search right most) into a single interface `findIndex`. Polymorphism is achieved through different arguments passed in by the caller.

PR #167, opened 2021-05-09
Check in 1854.py.

bunnygod:


1854.py
```diff
@@ -0,0 +1,16 @@
+class Solution:
+    def maximumPopulation(self, logs: List[List[int]]) -> int:
+        year_population = collections.defaultdict(int)
+        earliest_year = float(inf)
+
+        for birth, death in logs:
+            for i in range(birth, death):
```
mujingw, line 7: This inner `for` loop here doesn't scale well. In this particular problem, we don't have that many number of years (`1950 <= birth_i < death_i <= 2050`), so the range of `[birth, death)` is small enough to allow us to get away with this potential issue. 

In many other problems similar to this, using this [template](https://github.com/bunnygod/leetcode-python/pull/156#pullrequestreview-655100128) improves time complexity.

PR #168, opened 2021-05-09
Check in 67.py.

bunnygod:


67.py
```diff
@@ -0,0 +1,33 @@
+class Solution:
+    def addBinary(self, a: str, b: str) -> str:
+        a_ptr, b_ptr = len(a) - 1, len(b) - 1
+        carry = 0
+        queue = collections.deque()
+
+        while a_ptr >= 0 or b_ptr >= 0:
```
mujingw, line 7: You mixed up the following two structures in your code:

```python 
while i >= 0 or j >= 0:
    if i >= 0:
    # process digits in a

    if j >= 0:
    # process digits in b
```

```python
while i >= 0 and j >= 0:
    # process digits in both a and b

while i >= 0:
    # process leftover digits in a

while j >= 0:
    # process leftover digits in b
```

bunnygod: I think I understand what you are saying. Which template is more readable?

mujingw: I like the second one. It's more lengthy, but it's also more logical to handle the digits in three explicit loops.

The first one is shorter, but the `if` statements complicate the execution flow.

mujingw:
This problem is related to [LC-415](https://github.com/bunnygod/leetcode-python/pull/142#issue-628490422).

PR #169, opened 2021-05-11
Improved 1021.py.

bunnygod:


mujingw (approved):
LGTM. This is smarter and much more efficient than what I just wrote:

```python
def removeOuterParentheses(self, S: str) -> str:
    l, r = 0, 0
    primitives = []
    start = 0
    
    for i, ch in enumerate(S):
        if ch == '(':
            l += 1
        else:
            r += 1
            
        if l == r:
            primitives.append(S[start:i+1])
            start = i + 1
            
    return "".join([s[1:-1] for s in primitives])

PR #170, opened 2021-05-11
Check in 88.py.

amoogler:


88.py
```diff
@@ -0,0 +1,21 @@
...
+        i, j, k = m - 1, n - 1, m + n - 1
+
+        while i >= 0 and j >= 0:
+            if nums1[i] > nums2[j]:
+                nums1[k] = nums1[i]
+                k -= 1
+                i -= 1
+            else:
+                nums1[k] = nums2[j]
+                k -= 1
```
mujingw, line 15: We could extract line 11 and line 15 to the `while` loop level. Variable `k` means the next available position in the merged array. It makes sense to update it outside of the if-else blocks.

PR #171, opened 2021-05-11
Check in 501.py.

bunnygod:


501.py
```diff
@@ -0,0 +1,30 @@
...
+        counter = collections.defaultdict(int)
+        stack = []
+        pushleft(stack, root)
+
+        while stack:
+            node = stack.pop()
+            counter[node.val] += 1
+            pushleft(stack, node.right)
+
+        mode_count = max(counter.values())
```
mujingw, line 24: The mode of the numbers in this tree can be obtained while you traverse the tree. As soon as you update each `counter[node.val]`, you can compare this latest count with the current maximum count. Therefore, line 24 is not needed.

PR #172, opened 2021-05-12
Check in 1762.py.

bunnygod:


mujingw (approved):
LGTM. This is better than my approach using a monotonic stack:

```python
def findBuildings(self, heights: List[int]) -> List[int]:
    res = []
    s = []
    
    for i, h in enumerate(heights):
        while s and s[-1][1] <= h:
            s.pop()

        s.append((i, h))
        
    return [p[0] for p in s]

PR #174, opened 2021-05-13
Check in 821.py

bunnygod:


821.py
```diff
@@ -0,0 +1,19 @@
...
+        res = []
+
+        for idx, char in enumerate(s):
+            if char == c:
+                prev_idx = idx
+
+            res.append(idx - prev_idx)
+
+        prev_idx = float('inf')
+        for i in range(len(s) - 1, -1, -1):
```
mujingw, line 13: It looks better if you add a blank line between line 12 and line 13. This is just a recommendation; the style guide doesn't specify what to do in cases like this.

This implementation is pretty efficient because it can finish in just two passes, strictly. My implementation based on a dictionary has some repetition involved: 

```python
def shortestToChar(self, s: str, c: str) -> List[int]:
    d = defaultdict(list)
    res = []
    
    for i, ch in enumerate(s):
        d[ch].append(i)
        
    for i, ch in enumerate(s):
        res.append(min([abs(i - j) for j in d[c]]))
        
    return res

PR #175, opened 2021-05-13
Check in 1720.py.

bunnygod:


1720.py
```diff
@@ -0,0 +1,10 @@
+class Solution:
+    def decode(self, encoded: List[int], first: int) -> List[int]:
+        res = [first]
+        i = 0
```
mujingw, line 4: Variable `i` is not needed in this function. You only need to access the last element in `res`, which can be done via `res[-1]`.

PR #176, opened 2021-05-14
Check in 811.py.

bunnygod:


811.py
```diff
@@ -0,0 +1,19 @@
...
+        domain_count = collections.defaultdict(int)
+        res = []
+
+        for cpdomain in cpdomains:
+            parts = cpdomain.split()
+            count, domain = int(parts[0]), parts[1]
+
+            for i in range(len(domain) - 1, -1, -1):
+                if domain[i] == '.':
+                    domain_count[domain[i + 1 : len(domain)]] += count
```
mujingw, line 12: `A[i:]` is equivalent to `A[i:N]` where `N` is the length of `A`.

PR #177, opened 2021-05-14
Check in 572.py.

bunnygod:


mujingw (approved):
LGTM. Can you figure out the time complexity of this algorithm?

bunnygod:
> LGTM. Can you figure out the time complexity of this algorithm?

The time complexity is O(mn), m is the size of the tree and n is the size of the subtree.

mujingw:
> > LGTM. Can you figure out the time complexity of this algorithm?
> 
> The time complexity is O(mn), m is the size of the tree and n is the size of the subtree.

Can you provide the reasoning that leads to this conclusion?

bunnygod:
> > > LGTM. Can you figure out the time complexity of this algorithm?
> > 
> > 
> > The time complexity is O(mn), m is the size of the tree and n is the size of the subtree.
> 
> Can you provide the reasoning that leads to this conclusion?

Outer function isSubtree() needs to visit all nodes of the tree in worst case, while each visit we need to call isSameTree(). isSameTree() needs to visit all nodes of the subtree in the worst case.

mujingw:
> > > > LGTM. Can you figure out the time complexity of this algorithm?
> > > 
> > > 
> > > The time complexity is O(mn), m is the size of the tree and n is the size of the subtree.
> > 
> > 
> > Can you provide the reasoning that leads to this conclusion?
> 
> Outer function isSubtree() needs to visit all nodes of the tree in worst case, while each visit we need to call isSameTree(). isSameTree() needs to visit all nodes of the subtree in the worst case.

LGTM. Makes sense.

PR #178, opened 2021-05-14
Check in 973.py.

bunnygod:


973.py
```diff
@@ -0,0 +1,17 @@
+import heapq
+
+class Solution:
+    def kClosest(self, points: List[List[int]], k: int) -> List[List[int]]:
+        heap, res = [], []
+
+        for point in points:
```
mujingw, line 7: You could just say:

```python
for x, y in points:
# ...
# ...

File: 973.py [hunk 2]
```diff
@@ -0,0 +1,17 @@
...
+            x, y = point[0], point[1]
+            distance = pow(x, 2) + pow(y, 2)
+            heapq.heappush(heap, (distance, point))
+
+        elements = heapq.nsmallest(k, heap)
+
+        for e in elements:
```
mujingw, line 14: For a better readability, I would write:

```python
for distance, point in elements:
    res.append(point)

PR #179, opened 2021-05-15
Check in 1041.py.

bunnygod:


mujingw (commented):
For left turn, can `face + 3` be changed to `face - 1`? The meaning of `face + 3` took me quite a while to figure out.

bunnygod:
> For left turn, can `face + 3` be changed to `face - 1`? The meaning of `face + 3` took me quite a while to figure out.

Hmm..I think it is better to make it either clockwise or counterclockwise. The confusion can be eliminated by adding a few comments.

mujingw:
> > For left turn, can `face + 3` be changed to `face - 1`? The meaning of `face + 3` took me quite a while to figure out.
> 
> Hmm..I think it is better to make it either clockwise or counterclockwise. The confusion can be eliminated by adding a few comments.

OK. It makes more sense now since you mentioned clockwise.

PR #180, opened 2021-05-15
Check in 359.py.

bunnygod:


359.py
```diff
@@ -0,0 +1,23 @@
+class Logger:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.messages = collections.defaultdict(str)
```
mujingw, line 7: Two issues here:

1. Your implementation of `shouldPrintMessage` does not have a use case where a default value is needed from `self.messages`, so there is no need to create a `defaultdict` here.
2. The syntax of `defaultdict(some_type)` means that the values in your dictionary have a default value of type `some_type`. However, in your `shouldPrintMessage` function, the keys are of `str` type while the values are of `int` type. This dictionary definition is not consistent with its actual usages.

bunnygod: Another bad thing about this design is, it is not scalable at all, the space will keep increasing, I will need to pop out the old records.

mujingw (commented):
The implementation of this design is good now. However, there is a problem with this design: the log messages keep piling into your data structure, and they never get cleaned up. This is not designed for running in a production environment.

bunnygod:
> The implementation of this design is good now. However, there is a problem with this design: the log messages keep piling into your data structure, and they never get cleaned up. This is not designed for running in a production environment.

We can embed the expired message clean-up logic inside `shouldPrintMessage`, but this will slow down this hot-path routine. Instead, we can have an asynchronous process to clean up expired message entries in the table.

PR #181, opened 2021-05-15
Check in 362.py.

bunnygod:


362.py
```diff
@@ -0,0 +1,45 @@
...
+        self.total += 1
+
+
+    def getHits(self, timestamp: int) -> int:
+        """
+        Return the number of hits in the past 5 minutes.
+        @param timestamp - The current timestamp (in seconds granularity).
+        """
+
+        while self.hits:
```
mujingw, line 30: Good:
- You aggregated the count of each `timestamp`.

Not so good:
- You are using a linear search approach to get back to 300 seconds ago. Considering that the elements in the `deque` is already sorted, binary search should be used to locate the indexes of `timestamp - 300 + 1` and `timestamp` in the queue.
- You are discarding the old timestamps from `deque`. It is okay in this problem, but what if the requirement for `getHits()` becomes: given any timestamp (which could happen to be something that you have already discarded), return the total number of hits between `[timestamp - 300 + 1, timestamp]`?

bunnygod: For not so good parts:
1. This makes sense!
2. We design based on the given requirements. If we have to support `getHits()` queries at any time, we have to trade more storage for that. We can optimize how we gonna use storage, for example, put `[300s, timestamp]` to memory and the rest to disk.

bunnygod: Do you have some other good idea on point 2?

PR #182, opened 2021-05-15
Check in 1348.py.

bunnygod:


mujingw (commented):
I don't know why, but I never feel like sitting down and taking a closer look at this problem. It's been more than half a year, but I still don't have a submission for this problem.

PR #183, opened 2021-05-16
Check in 13.py.

bunnygod:


13.py
```diff
@@ -0,0 +1,21 @@
...
+    "C": 100,
+    "D": 500,
+    "M": 1000,
+}
+
+class Solution:
+    def romanToInt(self, s: str) -> int:
+        total = value_map[s[-1]]
+
+        for i in range(len(s) - 2, -1, -1):
```
mujingw, line 15: LGTM. A pythonic way to enumerate from the back would be:

```python
for idx, val in reversed(list(enumerate(A))):

mujingw (approved):

PR #184, opened 2021-05-16
Check in 1863.py.

bunnygod:


mujingw (approved):
LGTM. Line 7 and line 9 are essentially performing the choose and un-choose actions in a standard backtracking process.

PR #186, opened 2021-05-17
Check in 1859.py.

bunnygod:


mujingw (commented):
This implementation is quite good. Most people's implementation involves sorting, which has an `O(NlogN)` time complexity.

1859.py
```diff
@@ -0,0 +1,9 @@
+class Solution:
+    def sortSentence(self, s: str) -> str:
+        words = s.split()
+        res = [0] * len(words)
+
+        for word in words:
+            res[int(word[-1]) - 1] = word[:len(word) - 1]
```
mujingw, line 7: `A[:N - 1]` is equivalent to `A[:-1]` where `N` is the length of `A`.

PR #188, opened 2021-05-17
Check in 1396.py.

bunnygod:


1396.py
```diff
@@ -0,0 +1,25 @@
...
+        self.check_ins = dict()
+        self.total_times = collections.defaultdict(list)
+
+
+    def checkIn(self, id: int, stationName: str, t: int) -> None:
+        self.check_ins[id] = (stationName, t)
+
+
+    def checkOut(self, id: int, stationName: str, t: int) -> None:
+        self.total_times[(self.check_ins[id][0], stationName)].append(t - self.check_ins[id][1])
```
mujingw, line 13: I would assign `self.check_ins[id][0]` and `self.check_ins[id][1]` to two variables `last_check_in_station` and `last_check_in_time`, respectively, to improve readability.

PR #189, opened 2021-05-17
Check in 270.py.

bunnygod:


270.py
```diff
@@ -0,0 +1,37 @@
...
+#             prev_value = node.val
+#             pushleft(stack, node.right)
+
+#         return prev_value
+
+class Solution:
+    def closestValue(self, root: TreeNode, target: float) -> int:
+        closest = root.val
+
+        while root:
```
mujingw, line 33: LGTM. Space complexity has improved. You can also add a `if target == root.val` condition inside the loop to allow early return.

PR #190, opened 2021-05-18
Check in 703.py.

bunnygod:


mujingw (approved):
LGTM. Can you solve this problem using a tree map (or a balanced binary search tree) as the underlying data structure, too?

PR #191, opened 2021-05-19
Check in 1472.py.

bunnygod:


1472.py
```diff
@@ -0,0 +1,33 @@
...
+        self.forwards = []
+
+
+    def visit(self, url: str) -> None:
+        self.backs.append(url)
+        self.forwards.clear()
+
+
+    def back(self, steps: int) -> str:
+        while len(self.backs) > 1 and steps > 0:
```
mujingw, line 14: LGTM. It was hard for me to come up with a clean solution during the contest when I first saw this problem. My solution was always wrong with some cases. I guess they were related to the `1`s and `0`s in line 14 and line 22.

PR #192, opened 2021-05-19
Check in 1244.py.

bunnygod:


mujingw (approved):
LGTM. But keep in mind that `heapq.nlargest` is nothing fancy; it's `Equivalent to: sorted(iterable, key=key, reverse=True)[:n]`, as mentioned [here](https://docs.python.org/3/library/heapq.html#heapq.nlargest).

PR #193, opened 2021-05-20
Check in 1656.py.

bunnygod:


1656.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, n: int):
+        self.orders = [''] * n
+        self.order_count = 0
+
+    def insert(self, idKey: int, value: str) -> List[str]:
+        self.orders[idKey - 1] = value
+        next_orders = []
+
+        for order in self.orders[self.order_count:]:
+            if not order:
```
mujingw, line 12: I haven't solved this problem yet, but I see in many solutions that this `if` block can be avoided. Usually early return statements inside a loop are a result of not-so-well organized logic inside the loop.

PR #194, opened 2021-05-20
Check in 1337.py.

bunnygod:


1337.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def kWeakestRows(self, mat: List[List[int]], k: int) -> List[int]:
+        rows, res = [], []
+
+        for i in range(len(mat)):
+            rows.append((Counter(mat[i])[1], i))
+
+        rows = sorted(rows, key=lambda x : x[0])
```
mujingw, line 8: If the key is simply `x[0]`, you probably don't need to explicitly specify it in the `key=` argument. Sorting, by default, uses the first element in a `tuple` as the key.

1337.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def kWeakestRows(self, mat: List[List[int]], k: int) -> List[int]:
+        rows, res = [], []
+
+        for i in range(len(mat)):
+            rows.append((Counter(mat[i])[1], i))
```
mujingw, line 6: Instead of counting all the soldiers and civilians in `O(N)` time, the weakness can also be determined by finding the position (index) of the first civilian using binary search in `O(log N)` time (because `all the 1's will appear to the left of all the 0's in each row`).

```python
def kWeakestRows(self, mat: List[List[int]], k: int) -> List[int]:
    def bin_search(row):
        left, right = 0, len(row)
        
        while left < right:
            mid = (left + right) // 2
            
            if row[mid] == 1:
                left = mid + 1
            else:
                right = mid
                
        return left
    
    h = []
    res = []
    
    for i, row in enumerate(mat):
        heappush(h, (bin_search(row), i))
        
    while len(res) < k:
        res.append(heappop(h)[1])

    return res

PR #195, opened 2021-05-21
Check in 1845.py.

bunnygod:


1845.py
```diff
@@ -0,0 +1,18 @@
+class SeatManager:
+
+    def __init__(self, n: int):
+        self.seats = [i for i in range(1, n + 1)]
```
mujingw, line 4: This line is okay because `self.seats` is sorted and later we will want to `heappop` the smallest element anyway. However, adding one line of `heapq.heapify()` to explicitly heapify `self.seats` will make the code more readable.

PR #196, opened 2021-05-21
Check in 349.py.

bunnygod:


mujingw (approved):
LGTM. Your previous implementation based on sorting was quite good, too--I never thought about it. 

Ten years ago this was probably a great problem to be solved during an interview. It allows the interviewee to demonstrate that they know algorithms (sorting) and data structures (hash set) and how to use them to facilitate problem solving.

PR #197, opened 2021-05-22
Check in 139.py.

bunnygod:


139.py
```diff
@@ -1,10 +1,23 @@
...
         words = set(wordDict)
 
-        @lru_cache
-        def dfs(s: str):
-            return s in words or any(s[:i] in words and dfs(s[i:]) for i in range(1, len(s)))
+        dp = [False] * (len(s) + 1)
+        dp[0] = True
+
+        for i in range(1, len(s) + 1):
+            if any(dp[j] and s[j:i] in words for j in range(i)):
```
mujingw, line 20: This can be simplified (and improved) to:

```python
for end in range(1, N + 1):
    dp[end] = any(dp[start] and s[start:end] in words for start in range(end))

PR #198, opened 2021-05-23
Check in 1869.py

bunnygod:


1869.py
```diff
@@ -0,0 +1,20 @@
+class Solution:
```
mujingw, line 1: There is too much manual work involved in this solution. Here's what I did:
```python
def checkZeroOnes(self, s: str) -> bool:
    max_ones, max_zeros = 0, 0
    
    for k, g in groupby(s):
        if k == '1':
            max_ones = max(max_ones, len(list(g)))
        else:
            max_zeros = max(max_zeros, len(list(g)))
            
    return max_ones > max_zeros
```

You can take a look at [this part](https://docs.python.org/3/library/itertools.html#itertools.groupby) in the docs.

PR #199, opened 2021-05-23
Check in 110.py.

bunnygod:


mujingw (commented):
Looks like there are a lot of repetitiveness: for each node, you first get the depth of each child of this node (which requires going all the way down to the leaf nodes), and then you go through each child again (which also requires going all the way down to the leaf nodes) to check if they are balanced. 

In Python, we can conveniently return multiple values from a function, and we could take advantage of this to write something like below.

```python
def isBalanced(self, root: TreeNode) -> bool:
    def helper(root):
        if not root:
            return True, 0
        
        left_ok, left_depth = helper(root.left)
        right_ok, right_depth = helper(root.right)
        
        if not left_ok or not right_ok:
            return False, float('inf')
        
        if abs(left_depth - right_depth) > 1:
            return False, float('inf')
        else:
            return True, max(left_depth, right_depth) + 1
        
    return helper(root)[0]

mujingw:
I revisited this problem a few weeks ago and now I have a even clearer view to this problem: it is essentially a post-order traversal problem. Here is my simplified code:

```python
def isBalanced(self, root: Optional[TreeNode]) -> bool:
    return self.dfs(root)[1]
    
def dfs(self, node):
    if not node:
        return 0, True
    
    l_depth, l_balanced = self.dfs(node.left)
    r_depth, r_balanced = self.dfs(node.right)
    
    return max(l_depth, r_depth) + 1, abs(l_depth - r_depth) <= 1 and l_balanced and r_balanced

PR #200, opened 2021-05-23
Check in 716.py.

bunnygod:


mujingw (commented):
LGTM

716.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def top(self) -> int:
+        return self.stack[-1]
+
+
+    def peekMax(self) -> int:
+        return self.max_stack[-1]
+
+
+    def popMax(self) -> int:
```
mujingw, line 35: In this particular design, `popMax()` has a time complexity of `O(N)`. We can do better than this by _not_ using stacks or lists as the underlying data structure. This problem can be as difficult as the [LRU Cache](https://leetcode.com/problems/lru-cache/) problem.

PR #201, opened 2021-05-25
Check in 981.py.

bunnygod:


981.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def set(self, key: str, value: str, timestamp: int) -> None:
+        self.store[key].append((timestamp, value))
+
+
+    def get(self, key: str, timestamp: int) -> str:
+        values = self.store[key]
+        left, right = 0, len(values)
+
+        while left < right:
```
mujingw, line 18: LGTM. If permitted by the interviewer, you could also use the `bisect.bisect_right()` [function](https://docs.python.org/3/library/bisect.html#bisect.bisect_right) to replace your own binary search implementation.

PR #202, opened 2021-05-25
Check in 64.py.

bunnygod:


64.py
```diff
@@ -0,0 +1,17 @@
+class Solution:
+    def minPathSum(self, grid: List[List[int]]) -> int:
+        m, n = len(grid), len(grid[0])
```
mujingw, line 3: I'd like to use all upper case letters to name lengths of arrays or lists, because they are constants.

64.py
```diff
@@ -0,0 +1,17 @@
...
+        dp = [[0] * n for _ in range(m)]
+
+        for i in range(m):
```
mujingw, line 6: LGTM. You can also solve this problem using a top-down approach with memoization (i.e., `@lru_cache(None)`).

PR #203, opened 2021-05-26
Check in 128.py.

bunnygod:


128.py
```diff
@@ -0,0 +1,17 @@
...
+    def longestConsecutive(self, nums: List[int]) -> int:
+        longest = 0
+        nums_set = set(nums)
+
+        for num in nums:
+            if num - 1 not in nums_set:
+                curr_num = num
+                curr_count = 1
+
+                while curr_num + 1 in nums_set:
```
mujingw, line 11: Your solution probably has a time complexity of `O(N^2)`, but the problem asks for an `O(N)` one. 

`You must write an algorithm that runs in O(n) time.`

mujingw: Never mind. The `while` loop only starts when we meet a start of a streak. I thought it starts at every element in `nums`.

mujingw (approved):

mujingw:
When you study the Union Find data structure, you'll probably revisit this problem.

PR #204, opened 2021-05-26
Check in 380.py.

bunnygod:


380.py
```diff
@@ -0,0 +1,52 @@
+class RandomizedSet:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.dict = dict()
```
mujingw, line 7: Keywords like `dict` and `list` are not good choices for variable names.

380.py
```diff
@@ -0,0 +1,52 @@
...
+        if val in self.dict:
+            return False
+
+        self.dict[val] = len(self.list)
+        self.list.append(val)
+
+        return True
+
+
+    def remove(self, val: int) -> bool:
```
mujingw, line 24: LGTM. `insert()` and `getRandom()` are trivial; the most difficult part of this problem is to come up with this "remove the last and overwrite something in the middle" trick for `remove()`.

PR #205, opened 2021-05-26
Check in 707.py.

bunnygod:


707.py
```diff
@@ -0,0 +1,91 @@
...
+        self.val = val
+        self.next = None
+
+class MyLinkedList:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.pre_head = Node(0)
```
mujingw, line 13: This is an annoying problem, but as long as you have this "dummy" `pre_head` node, you have already avoided (probably) 90% of the bugs related to linked list problems.

707.py
```diff
@@ -0,0 +1,91 @@
...
+    def addAtHead(self, val: int) -> None:
+        """
+        Add a node of value val before the first element of the linked list. After the insertion, the new node will be the first node of the linked list.
+        """
+        head = Node(val)
+        head.next = self.pre_head.next
+        self.pre_head.next = head
+
+
+    def addAtTail(self, val: int) -> None:
```
mujingw, line 39: This `addAtTail()` operation can be done in `O(1)` time if you maintain a tail `Node` in your `MyLinkedList` class.

PR #206, opened 2021-05-27
Check in 290.py.

bunnygod:


290.py
```diff
@@ -0,0 +1,19 @@
...
+
+        str_to_pattern = dict()
+        pattern_to_str = dict()
+
+        for idx, (s, p) in enumerate(zip(strs, pattern)):
+            if (s in str_to_pattern and str_to_pattern[s] != pattern[idx]) or \
+               (p in pattern_to_str and pattern_to_str[p] != strs[idx]):
+                return False
+
+            str_to_pattern[s] = pattern[idx]
```
mujingw, line 16: In this `for` loop, you already have `p` and `s` that is equivalent of `pattern[idx]` and `strs[idx]`, respectively. I would rather use `p` and `s` here.

290.py
```diff
@@ -0,0 +1,19 @@
+class Solution:
+    def wordPattern(self, pattern: str, s: str) -> bool:
+        strs = s.split()
```
mujingw, line 3: `strs` is not descriptive. An abbreviated variable name like `strs` doesn't fit with a name like `pattern` in the same function. Let's use something like `words` for consistency.

290.py
```diff
@@ -0,0 +1,19 @@
...
+
+        if len(pattern) != len(strs):
+            return False
+
+        str_to_pattern = dict()
+        pattern_to_str = dict()
+
+        for idx, (s, p) in enumerate(zip(strs, pattern)):
+            if (s in str_to_pattern and str_to_pattern[s] != pattern[idx]) or \
```
mujingw, line 12: I would break this into two `if` statements:

```python
def wordPattern(self, pattern: str, s: str) -> bool:
    p2w = {}
    w2p = {}
    words = s.split()
    
    if len(pattern) != len(words):
        return False
    
    for p, word in zip(pattern, words):
        if p in p2w and p2w[p] != word:
            return False
        
        if word in w2p and w2p[word] != p:
            return False
        
        if p not in p2w:
            p2w[p] = word
            
        if word not in w2p:
            w2p[word] = p
            
    return True

mujingw (commented):
LGTM

PR #207, opened 2021-05-27
Check in 1756.py.

bunnygod:


1756.py
```diff
@@ -0,0 +1,25 @@
+class MRUQueue:
```
mujingw, line 1: Your current design has the same time complexity as mine, which is `O(N)`. However, your actual implementation is unnecessarily complicated. Here's what I wrote:

```python
class MRUQueue:

    def __init__(self, n: int):
        self.q = [x for x in range(1, 1 + n)]
        
    def fetch(self, k: int) -> int:
        res = self.q[k-1]
        self.q = self.q[:k-1] + self.q[k:] + [res]
        
        return res
```

PR #208, opened 2021-05-28
Check in 211.py

bunnygod:


211.py
```diff
@@ -0,0 +1,46 @@
...
+            curr_node = curr_node.children[c]
+
+        curr_node.is_word = True
+
+
+    def search(self, word: str) -> bool:
+        def search_backtrack(word: str, node: TrieNode) -> bool:
+            for i, c in enumerate(word):
+                if c not in node.children:
+                    if c == '.' and any(search_backtrack(word[i + 1:], node.children[x]) for x in node.children):
```
mujingw, line 32: You don't need `x` here. You can just say:

```python
any(search_backtrack(word[i+1:], next_node) for next_node in node.children.values())

File: 211.py [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+                        return True
```
mujingw, line 33: The logic can be simplified a little bit:

```python
if c not in node.children:
    if c == '.':
        return any(search_backtrack(word[i+1:], next_node) for next_node in node.children.values())
    else:
        return False
else:
    node = node.children[c]

PR #209, opened 2021-05-29
Check in 1381.py.

bunnygod:


mujingw (commented):
This implementation is OK, but you probably want to revisit this problem once you learn the [prefix sum array](https://labuladong.github.io/algo/4/31/108/), [difference array](https://labuladong.github.io/algo/4/31/109/), or sweep line techniques.

mujingw (approved):

PR #211, opened 2021-05-30
Check in 286.py.

bunnygod:


286.py
```diff
@@ -0,0 +1,64 @@
...
+
+        DIRS = ((1, 0), (-1, 0), (0, 1), (0, -1))
+
+        while queue:
+            row, col = queue.popleft()
+
+            for dr, dc in DIRS:
+                r, c = row + dr, col + dc
+
+                if r < 0 or r >= M or c < 0 or c >= N or rooms[r][c] != EMPTY:
```
mujingw, line 60: It's probably better to invert the logic and use the pythonic way for range checks:

```python
if 0 <= r < R and 0 <= c < C and rooms[r][c] == EMPTY:
    # do something 
```

286.py
```diff
@@ -0,0 +1,64 @@
...
+                    continue
+
+                rooms[r][c] = rooms[row][col] + 1
```
mujingw, line 63: Make sure you understand what this line does and why we don't need a `seen` variable (to accompany `queue`) in this problem's solutions.

```python
rooms[next_row][next_col] = rooms[curr_row][curr_col] + 1

**Comment by mujingw** [comment]
Pasting mine for your reference:

```python
from collections import deque


class Solution:
    def wallsAndGates(self, rooms: List[List[int]]) -> None:
        """
        Do not return anything, modify rooms in-place instead.
        """
        R, C = len(rooms), len(rooms[0])
        DIR = ((0, 1), (0, -1), (-1, 0), (1, 0))
        EMPTY, GATE = 2147483647, 0
        q = deque([])
        
        for r in range(R):
            for c in range(C):
                if rooms[r][c] == GATE:
                    q.append((r, c, 0))
                    
        while q:
            cx, cy, dist = q.popleft()
            
            for dx, dy in DIR:
                nx, ny = cx + dx, cy + dy
                
                if 0 <= nx < R and 0 <= ny < C and rooms[nx][ny] > dist + 1:
                    rooms[nx][ny] = dist + 1
                    q.append((nx, ny, dist + 1))

File: 286.py (hunk 1 above)
Comment (bunnygod) [reply 286.py:60]: I kind of disagree, usually bail early and save an indentation is the way to go.

**Comment by mujingw** [comment]
There are two ways to solve this problem.

1. From each gate, populate and **_update_** each initially-empty cell with the shortest distance (this is what you implemented this time).
2. From each initially-empty cell, find the closest gate to this cell.

I suggest you solve this problem again using the second approach and compare their advantages and disadvantages. For more practice and a deeper understanding of this type of problems, take a look at [LC-542](https://leetcode.com/problems/01-matrix/).

File: 286.py (hunk 1 above)
Comment (mujingw) [reply 286.py]: This is not a fail early situation. This block is essentially doing this:

```python
if not A:
    continue
else:
    # execute S1
    # execute S2
```
Isn't it equivalent of this?

```python
if A:
    # execute S1
    # execute S2
```
Indentation-wise, there is no difference, either. I'll leave it to you to decide which one is less fragmented and more readable.

mujingw:
> There are two ways to solve this problem.
> 
> 1. From each gate, populate and **_update_** each initially-empty cell with the shortest distance (this is what you implemented this time).
> 2. From each initially-empty cell, find the closest gate to this cell.
> 
> I suggest you solve this problem again using the second approach and compare their advantages and disadvantages. For more practice and a deeper understanding of this type of problems, take a look at [LC-542](https://leetcode.com/problems/01-matrix/).

Never mind. From the code that you commented out, I see that you have already tried the second approach I mentioned above and got a TLE. I got a TLE for that approach too. This is still a good practice regardless.

mujingw:
> Pasting mine for your reference:
> 
> ```python
> from collections import deque
> 
> 
> class Solution:
>     def wallsAndGates(self, rooms: List[List[int]]) -> None:
>         """
>         Do not return anything, modify rooms in-place instead.
>         """
>         R, C = len(rooms), len(rooms[0])
>         DIR = ((0, 1), (0, -1), (-1, 0), (1, 0))
>         EMPTY, GATE = 2147483647, 0
>         q = deque([])
>         
>         for r in range(R):
>             for c in range(C):
>                 if rooms[r][c] == GATE:
>                     q.append((r, c, 0))
>                     
>         while q:
>             cx, cy, dist = q.popleft()
>             
>             for dx, dy in DIR:
>                 nx, ny = cx + dx, cy + dy
>                 
>                 if 0 <= nx < R and 0 <= ny < C and rooms[nx][ny] > dist + 1:
>                     rooms[nx][ny] = dist + 1
>                     q.append((nx, ny, dist + 1))
> ```

`rooms[nx][ny] > dist + 1` is not needed in my code. I could have just said `rooms[nx][ny] == EMPTY`. In this problem, it is impossible to see `(nx, ny)` for a second time with a smaller `dist + 1`.

PR #212, opened 2021-05-30
Check in 1880.py.

bunnygod:


mujingw (approved):
LGTM, although the `for` loop can be written as this:

```python

for ch in word[::-1]:
    v = (ord(ch) - ord('a')) * base
    value += v
    base *= 10
```

During a contest, I suggest that we use as many built-in functions as we can:

```python

def isSumEqual(self, f: str, s: str, t: str) -> bool:
    def w2num(word):
        res = []
        
        for ch in word:
            res.append(str(ord(ch) - ord('a')))
        
        return int("".join(res))
    
    return w2num(f) + w2num(s) == w2num(t)

PR #213, opened 2021-05-30
Check in 1881.py.

bunnygod:


1881.py
```diff
@@ -0,0 +1,17 @@
...
+                if x < int(n_list[i]):
+                    n_list.insert(i, str(x))
+                    return ''.join(n_list)
+        else:
+            for i in range(0, len(n_list)):
+                if x > int(n_list[i]):
+                    n_list.insert(i, str(x))
+                    return ''.join(n_list)
+
+        n_list.insert(len(n_list), str(x))
```
mujingw, line 16: If inserting at the end of the list, we can also say:

```python
n_list.append(str(x))

mujingw (approved):
LGTM. Overall this is a clean implementation. It is much better than mine, which failed three times during the contest.

PR #215, opened 2021-05-31
Check in 621.py.

bunnygod:


621.py
```diff
@@ -0,0 +1,20 @@
+class Solution:
+    def leastInterval(self, tasks: List[str], n: int) -> int:
+        task_freqs = Counter(tasks)
+        sorted_freqs = sorted(task_freqs.values(), reverse=True)
```
mujingw, line 4: The code between line 4 and line 12 is too verbose and doesn't have a good performance due to the sorting part. I would write it this way:

```python
max_freq = max(task_freq.values())
max_count = task_freq.values().count(max_freq)
```

621.py
```diff
@@ -0,0 +1,20 @@
...
+
+        max_freq, max_count = sorted_freqs[0], 1
+
+        for freq in sorted_freqs[1:]:
+            if freq == max_freq:
+                max_count += 1
+            else:
+                break
+
+        part_count = max_freq - 1
```
mujingw, line 14: The names of `part_count` and `part_length` don't really make sense to me. They need to be more descriptive.

621.py
```diff
@@ -1,20 +1,13 @@
...
-            else:
-                break
-
-        part_count = max_freq - 1
-        part_length = n - (max_count - 1)
-        empty_slots = part_count * part_length
-        left_task_slots = len(tasks) - max_freq * max_count
-        idles = max(0, empty_slots - left_task_slots)
+        max_freq = max(task_freqs.values())
+        max_count = list(task_freqs.values()).count(max_freq)
```
mujingw, line 5: `max_count` should really be `max_freq_count`. `max_count` alone doesn't mean anything, i.e., maximum count of what?

621.py
```diff
@@ -1,20 +1,13 @@
...
+
+        interval_count = max_freq - 1
```
mujingw, line 7: The code between line 7 and line 11 is very concise, but the mathematical reasoning behind it is not clear. We definitely need to revisit this problem at some point later.

PR #216, opened 2021-05-31
Check in 513.py

bunnygod:


mujingw (commented):
Same issue as mentioned [here](https://github.com/bunnygod/leetcode-python/pull/41#discussion_r608209764).

mujingw:
One more thing: although this problem can be solved by BFS, it is extremely slow. For problems with this nature (deepest, longest, etc.), using DFS is more preferable.

mujingw:
> One more thing: although this problem can be solved by BFS, ~it is extremely slow~. For problems with this nature (deepest, longest, etc.), using DFS is more preferable.

After a second thought, I realized that the time complexity of the BFS approach is the same as the DFS', because either algorithm has to visit all nodes in the tree. In terms of space complexity, it depends on the shape of the tree. The tree can be very flat but very wide; it can also be very tall (vertically) but very narrow (horizontally).

PR #217, opened 2021-06-01
Check in 146.py.

bunnygod:


mujingw (approved):
LGTM. `_move_to_head` and `_pop_tail` helped keep the implementation clean.

PR #218, opened 2021-06-01
Check in 1448.py.

bunnygod:


1448.py
```diff
@@ -0,0 +1,22 @@
...
+#         self.left = left
+#         self.right = right
+class Solution:
+    def goodNodes(self, root: TreeNode) -> int:
+        def dfs(node: TreeNode, curr_max: int, res: list):
+            if node.val >= curr_max:
+                res[0] += 1
+                curr_max = max(node.val, curr_max)
+
+            if node.left:
```
mujingw, line 14: I would delegate the nullity check to the child level, as mentioned [here](https://github.com/bunnygod/leetcode-python/pull/4#issuecomment-810762033).

Here's my implementation for your reference:

```python
def goodNodes(self, root: TreeNode) -> int:
    def dfs(res, node, mx):
        if not node:
            return
    
        if node.val >= mx:
            res[0] += 1
            mx = node.val            
    
        dfs(res, node.left, mx)
        dfs(res, node.right, mx)

    res = [0]
    dfs(res, root, float('-inf'))
    
    return res[0]

PR #219, opened 2021-06-02
Check in 708.py.

bunnygod:


708.py
```diff
@@ -0,0 +1,32 @@
...
+    def insert(self, head: 'Node', insertVal: int) -> 'Node':
+        node = Node(insertVal)
+
+        if not head:
+            node.next = node
+            return node
+
+        prev, curr = head, head.next
+
+        while prev.next != head:
```
mujingw, line 19: Having `prev.next != head` here and avoiding `while True` is much better than having the third `if` statement in the implementation below:

```python
while True:
    # normal case
    if prev.val <= insert_val <= curr.val:
        break
    
    # peak, valley
    if prev.val > curr.val and (insert_val > prev.val or insert_val < curr.val):
        break
     
    prev, curr = prev.next, curr.next

    # all nodes have the same value
    if prev == head:
        break

mujingw (approved):
LGTM

PR #220, opened 2021-06-02
Check in 1518.py.

bunnygod:


1518.py
```diff
@@ -0,0 +1,16 @@
+class Solution:
+    def numWaterBottles(self, numBottles: int, numExchange: int) -> int:
+        total, left_empty, to_drink = 0, 0, numBottles
```
mujingw, line 3: Function parameters provided by LeetCode don't follow Python naming conventions. We should rename `numBottles` and `numExchange` such that they are consistent with the other variables in the solution.

1518.py
```diff
@@ -0,0 +1,16 @@
...
+
+        while to_drink != 0:
+            # Drink and count.
+            drunk = to_drink
+            total += drunk
+
+            # Compute how many new drinks we can get.
```
mujingw, line 10: "Compute" is too generic. How about "update total number of drinks available for the next round"?

1518.py
```diff
@@ -0,0 +1,16 @@
...
+            to_drink = (drunk + left_empty) // numExchange
+
+            # Compute how many empty bottles after exchange.
```
mujingw, line 13: Similarly, how about "update number of empty bottles after exchange"?

mujingw:
LGTM. Here's my version:

```python
def numWaterBottles(self, bottles: int, exchange: int) -> int:
    total = 0
    
    while bottles >= exchange:
        can_exchange, left_over = divmod(bottles, exchange)
        total += (bottles - left_over)
        bottles = can_exchange + left_over
        
    return total + bottles % exchange

PR #222, opened 2021-06-04
Check in 243.py.

bunnygod:


mujingw (approved):
This is a very C-style implementation. You declare two index variables with dummy values first and then do some processing. The good thing is that its time complexity is great--`O(N)` and finishes one-pass; the down side is that it's hard to read.

My implementation below is more compact but goes through the list of words more than once and can potentially lead to a `O(N^2)` time complexity (when one half of the words in `words` is `word1` and the other half of words in `words` is `word2`).

```python
def shortestDistance(self, words, word1, word2):
    lookup = defaultdict(list)
    res = float("inf")

    for i, word in enumerate(words):
        lookup[word].append(i)

    for i in lookup[word1]:
        for j in lookup[word2]:
            res = min(res, abs(i-j))
            
    return res

mujingw:
There is a good `O(N)` time and `O(1)` space implementation posted [here](https://leetcode.com/problems/shortest-word-distance/discuss/66947/Python-solution-O(n)-time-O(1)-space), which is more readable. 

```python
def shortestDistance(self, words, word1, word2):
    size = len(words)
    index1, index2 = size, size
    ans = size
    
    for i in range(size):
        if words[i] == word1:
            index1 = i
            ans = min(ans, abs(index1-index2))
        elif words[i] == word2:
            index2 = i
            ans = min(ans, abs(index1-index2))

    return ans

PR #223, opened 2021-06-05
Check in 7.py.

bunnygod:


mujingw (approved):
To solve this problem in a mathematical way (which could be the original intention of asking this problem), this is probably as clean as it gets. Handling integer overflow during processing is another story though.

There are idioms in C and C++ to detect overflow on signed integers, which are mostly based on the change of the sign of the integer when overflow occurs. In Python this technique cannot be applied because integers are [implemented and stored differently.](https://docs.python.org/3/c-api/long.html#integer-objects)

On the other hand, since [integer overflow never happens in Python](https://docs.python.org/3/library/exceptions.html#OverflowError), artificially handling integer overflow (i.e., handling it just for the sake of handling it) is a moot point.

bunnygod:
> To solve this problem in a mathematical way (which could be the original intention of asking this problem), this is probably as clean as it gets. Handling integer overflow during processing is another story though.
> 
> There are idioms in C and C++ to detect overflow on signed integers, which are mostly based on the change of the sign of the integer when overflow occurs. In Python, this technique cannot be applied because integers are [implemented and stored differently.](https://docs.python.org/3/c-api/long.html#integer-objects)
> 
> On the other hand, since [integer overflow never happens in Python](https://docs.python.org/3/library/exceptions.html#OverflowError), artificially handling integer overflow (i.e., handling it just for the sake of handling it) is a moot point.

We need to re-do this problem with overflow handling, possibly with C/C++. This problem is a good one if you interview for a position with a focus on system-level programming.

mujingw:
> > To solve this problem in a mathematical way (which could be the original intention of asking this problem), this is probably as clean as it gets. Handling integer overflow during processing is another story though.
> > There are idioms in C and C++ to detect overflow on signed integers, which are mostly based on the change of the sign of the integer when overflow occurs. In Python, this technique cannot be applied because integers are [implemented and stored differently.](https://docs.python.org/3/c-api/long.html#integer-objects)
> > On the other hand, since [integer overflow never happens in Python](https://docs.python.org/3/library/exceptions.html#OverflowError), artificially handling integer overflow (i.e., handling it just for the sake of handling it) is a moot point.
> 
> We need to re-do this problem with overflow handling, possibly with C/C++. This problem is a good one if you interview for a position with a focus on system-level programming.

Sure.

PR #226, opened 2021-06-06
Check in 1887.py.

bunnygod:


1887.py
```diff
@@ -0,0 +1,20 @@
...
+        largest = sorted_nums[0]
+        smallest = sorted_nums[len(nums) - 1]
+        L = len(set(nums))
+
+        for num in sorted_nums:
+
+            if num == smallest:
+                break
+            elif num != largest:
+                L -= 1
```
mujingw, line 15: If it's capitalized it's, by convention, a constant. You should not modify your constant.

1887.py
```diff
@@ -0,0 +1,20 @@
+class Solution:
+    def reductionOperations(self, nums: List[int]) -> int:
+        res = 0
+
+        sorted_nums = sorted(nums, reverse=True)
+        largest = sorted_nums[0]
+        smallest = sorted_nums[len(nums) - 1]
```
mujingw, line 7: You need to get used to the pythonic `nums[-1]` notation.

mujingw (commented):
I'm pasting mine for your reference:

```python
def reductionOperations(self, nums: List[int]) -> int:
    c = Counter(nums)
    res = 0
    accumulated = 0
      
    for val in sorted(list(c.keys()), reverse=True)[:-1]:
        res += (accumulated + c[val])
        accumulated += c[val]
        
    return res

PR #227, opened 2021-06-06
Check in 73.py.

bunnygod:


mujingw (commented):
Here's my implementation based on the same idea as yours:

```python
def setZeroes(self, matrix: List[List[int]]) -> None:
    """
    Do not return anything, modify matrix in-place instead.
    """
    R, C = len(matrix), len(matrix[0])
    should_clear_col0 = any(matrix[r][0] == 0 for r in range(R))
    should_clear_row0 = any(matrix[0][c] == 0 for c in range(C))
    
    for i in range(R):
        for j in range(C):
            if matrix[i][j] == 0:
                matrix[0][j] = 0
                matrix[i][0] = 0
    
    for i in range(1, R):
        for j in range(1, C):
            if matrix[0][j] == 0 or matrix[i][0] == 0:
                matrix[i][j] = 0
    
    if should_clear_col0:
        for i in range(R):
            matrix[i][0] = 0
    
    if should_clear_row0:
        for j in range(C):
            matrix[0][j] = 0
    
```
```
Runtime: 116 ms, faster than 99.21% of Python3 online submissions for Set Matrix Zeroes.
Memory Usage: 15.3 MB, less than 11.55% of Python3 online submissions for Set Matrix Zeroes.
```

PR #229, opened 2021-06-07
Check in 236.py.

bunnygod:


236.py
```diff
@@ -0,0 +1,24 @@
...
+#         self.right = None
+
+class Solution:
+    def lowestCommonAncestor(self, root: 'TreeNode', p: 'TreeNode', q: 'TreeNode') -> 'TreeNode':
+        if root == p or root == q:
+            return root
+
+        left, right = None, None
+
+        if root.left:
```
mujingw, line 15: If your code keeps repeating the same problem I raised over and over again, what is the point of doing this code review? What have you learned from doing [LC-111](https://github.com/bunnygod/leetcode-python/pull/4#issuecomment-810762033) and [LC-1448](https://github.com/bunnygod/leetcode-python/pull/218#discussion_r642782024) and reading the code review comments? It's okay if you repeat the same issue only once, but repeating it twice is not something a good and serious learner should display.  

Code review is not a drama that wastes everyone's time; it is a forum for everyone to learn something and gradually improve over time. I feel like I, as a reviewer, am spending more time on thinking the problems and learning from your code than you are, and as a result I am getting more out of this code review process.

mujingw: To be clear, this is not just a coding style issue. In fact, what you wrote here is incorrect--a simple test case with a `None` root will lead to a crash in your code (`AttributeError` caused by `root.left` and `root.right`). Your code passed the LeetCode online judge because they don't have such a test case and they mentioned `The number of nodes in the tree is in the range [2, 105]`. But from an engineering perspective, you don't really want the reliability of your code to depend on the input, do you?

More formally, this implementation is incorrect because it lacks a base case: any node in a tree could be `None` and you clearly didn't handle this case.

I keep emphasizing this point because what you are doing here is exactly what keeps you away from consistently writing correct recursions, which always consist of [two parts](https://en.wikipedia.org/wiki/Recursion#Formal_definitions): 

- A simple base case (or cases) — a terminating scenario that does not use recursion to produce an answer
- A recursive step — a set of rules that reduces all successive cases toward the base case

If you have a clearer understanding of the two properties above, you'll eventually be more comfortable writing recursive implementations (including but not limited to DFS and backtracking).

Consistently writing correct code sometimes requires mathematical considerations, as in recursion and binary search. My hope is that you write your code because you know it's correct with confidence backed by theories in Discrete Math, not because the debugger or LeetCode results show that it _seems_ to be correct (i.e., debugger-oriented programming).

PR #230, opened 2021-06-08
Check in BFS solution for 200.py.

bunnygod:


mujingw (commented):
Three issues in this implementation: 

1. In your DFS implementation (of this exact problem) you remembered to use `dr`, `dc` to simplify your "visit" logic, but in this BFS implementation you rolled back to a hard-coding style.
2. [Here](https://github.com/bunnygod/leetcode-python/pull/211#discussion_r642087209) you were aware of the number of indentation levels, but in this BFS implementation, have you thought about extracting your BFS logic into a standalone function to avoid the indentation problem? As you can see, the code gets ugly when the `while` loop starts. 
3. In addition, having a standalone BFS helper function improves the modularity of your code because it is called by every cell `grid[i][j]`.

Regardless, this is still a good exercise because at least I can tell this is your own code. Making mistakes like these is no big deal; you never learn or fully digest the learning materials until you make mistakes and correct them.

PR #231, opened 2021-06-08
Check in 54.py.

bunnygod:


mujingw (commented):
You are confused with the naming conventions: all capitalized names are used for constants (e.g., `final` in Java, `const` in C++, `readonly` in C#). In this `spiralOrder` function, `R_BEG`, `R_END`, `C_BEG`, and `C_END` are all changing in the rest of the code, so they are not constants.

This implementation itself is good. The key here is to maintain the loop invariant `r_beg <= r_end and c_beg <= c_end` and make sure it remains `True` inside the loop (i.e., line 21 and line 27).

mujingw (commented):
The current implementation is an acceptable one for junior engineers--it's clean and logical. However, for senior folks like us, we do things more elegantly:

```python3
def spiralOrder(self, matrix: List[List[int]]) -> List[int]:
    R, C = len(matrix), len(matrix[0])
    DIR = ((0, 1), (1, 0), (0, -1), (-1, 0))
    dir_idx = 0
    dx, dy = DIR[dir_idx % len(DIR)]
    cx, cy = 0, -1
    visited = set()
    res = []
    
    while len(res) < R * C:
        nx, ny = cx + dx, cy + dy
        
        if (0 <= nx < R and 0 <= ny < C) and (nx, ny) not in visited:
            visited.add((nx, ny))
            res.append(matrix[nx][ny])
            cx, cy = nx, ny
        else:
            dir_idx += 1
            dx, dy = DIR[dir_idx % len(DIR)]
    
    return res
```

What's nice about this new implementation is that it avoids lots of manual (hardcoded) checks and adjustments inside the `while` loop body. State (x and y coordinates) transitions occur automatically in a predefined, well structured way.

PR #232, opened 2021-06-09
Check in 542.py.

bunnygod:


mujingw (commented):
This BFS implementation is better structured than [yesterday's](https://github.com/bunnygod/leetcode-python/pull/230/commits/3434f37f3743e3aa4924e67aa86ea15f41cef4d2). However, make sure you understand:

1. Why the implementation that you commented out resulted in TLE.
2. The difference (in terms of time complexity) between your AC version and your TLE version.

PR #234, opened 2021-06-09
Check in 994.py

bunnygod:


mujingw (commented):
This implementation shows continued improvements that we see in the last couple of days, which is good.

994.py
```diff
@@ -0,0 +1,40 @@
...
+        if fresh_oranges == 0:
+            return 0
+
+        minute = 0
+        DIRS = ((1, 0), (-1, 0), (0, 1), (0, -1))
+
+        while queue:
+            queue_length = len(queue)
+
+            for _ in range(queue_length):
```
mujingw, line 24: I can see this is the template you used when solving the tree-based BFS problems, and you are able to apply it here.

994.py
```diff
@@ -0,0 +1,40 @@
...
+                r, c = queue.popleft()
+
+                for dr, dc in DIRS:
+                    new_r, new_c = r + dr, c + dc
+
+                    if new_r < 0 or new_r >= R or new_c < 0 or new_c >= C or \
+                        grid[new_r][new_c] != 1:
+                        continue
+
+                    grid[new_r][new_c] = 2
```
mujingw, line 34: Your BFS template is good, for now, but it still lacks one key component: it doesn't track which cells or nodes it has already "processed" (Note: By "processed", I am intentionally being vague in my word choice here, because I want you to figure out what to do in your code).

Of course, when solving those tree-based and grid-based problems, you haven't really been in a situation where you have to deal with "processed" elements:

1. There is no cycle in trees, so you never needed to take care of "processed" nodes
2. You can modify the cells inside a grid, which indirectly help you avoid repetitive work

But what if you are in one of these two situations?

1. You are not allowed to modify the grid
2. You are dealing with a generic graph (not just trees) with cycles

994.py
bunnygod: Sure, I plan to do that soon.

994.py
bunnygod: I can allocate a tracking grid to maintain "visited" status, is that what you're trying to imply?

In this particular problem, yea, we don't have to do that, `grid[new_r][new_c] != 1` will take care of it.

mujingw: Having an extra grid works (as you suggested above), but there are two issues with this approach:

1. It only works when you are dealing with a grid-related problem
2. It occupies too much space, which has the same size as the input grid

I was expecting your answer to be: maintaining a `set` of grid coordinates (represented as tuples) for "visited" cells, which is a more generic and more space efficient approach. This way, you are only storing cells that are relevant to your traversal whereas in the extra grid approach you will have to allocate space for all cells.

More generally, what we really need to do is to maintain a `set` of "visited" vertices in the graph (this graph just happens to be a grid in this problem).

PR #235, opened 2021-06-11
Check in 463.py.

bunnygod:


463.py
```diff
@@ -0,0 +1,21 @@
+class Solution:
+    def islandPerimeter(self, grid: List[List[int]]) -> int:
+        R, C = len(grid), len(grid[0])
+        perimeter = 0
+
+        for i in range(R):
+            for j in range(C):
+                if grid[i][j] == 1:
+                    if i - 1 < 0 or (i - 1 >= 0 and grid[i - 1][j]) == 0:
```
mujingw, line 9: There is no need to write `i - 1 >= 0` after `i - 1 < 0` due to logical operators' [short circuiting](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not) behavior in Python.

Also, `i - 1 < 0` can be replaced by `i == 0` for better readability.

mujingw:
For this kind of grid-related problems, I find that using a `DIR` tuple greatly improves readability and maintainability.

```python
def islandPerimeter(self, grid: List[List[int]]) -> int:
    R, C = len(grid), len(grid[0])
    DIR = ((0, 1), (0, -1), (1, 0), (-1, 0))
    res = 0
    
    for r, c in product(range(R), range(C)):
        if grid[r][c] == 0:
            continue
            
        for dx, dy in DIR:
            nx, ny = r + dx, c + dy

            if not (0 <= nx < R and 0 <= ny < C):
                res += 1
            elif grid[nx][ny] == 0:
                res += 1

    return res

PR #237, opened 2021-06-12
Check in 490.py.

bunnygod:


490.py
```diff
@@ -0,0 +1,73 @@
...
+# DFS solution.
+
+
+class Solution:
+    def hasPath(self, maze: List[List[int]], start: List[int], destination: List[int]) -> bool:
+        R, C = len(maze), len(maze[0])
+        visited = [[False] * C for _ in range(R)]
+        DIRS = ((1, 0), (-1, 0), (0, 1), (0, -1))
+        queue = deque([start])
+        visited[start[0]][start[1]] = True
```
mujingw, line 54: I strongly object to this `visited` name here. Technically, when you mark a cell to `True`, you have **_not_** "visited" this cell yet; you just discovered this cell and put it into a queue. On Wikipedia, the BFS pseudo code uses `discovered` to name this variable and the LeetCode community tend to use `seen`, both of which are okay with me.

490.py
```diff
@@ -0,0 +1,73 @@
...
+        while queue:
+            sr, sc = queue.popleft()
+
+            if sr == destination[0] and sc == destination[1]:
+                return True
+
+            for dr, dc in DIRS:
+                new_r, new_c = sr + dr, sc + dc
+
+                while new_r >= 0 and new_r < R and new_c >= 0 and new_c < C and maze[new_r][new_c] == 0:
```
mujingw, line 65: Same problem as mentioned [here](https://github.com/bunnygod/leetcode-python/pull/236#discussion_r659317346).

490.py
```diff
@@ -0,0 +1,73 @@
...
+class Solution:
+    def hasPath(self, maze: List[List[int]], start: List[int], destination: List[int]) -> bool:
+        R, C = len(maze), len(maze[0])
+        visited = [[False] * C for _ in range(R)]
+        DIRS = ((1, 0), (-1, 0), (0, 1), (0, -1))
+        queue = deque([start])
+        visited[start[0]][start[1]] = True
+
+        while queue:
+            sr, sc = queue.popleft()
```
mujingw, line 57: What do `sr` and `sc` mean here? Starting row and starting column?

490.py
```diff
@@ -0,0 +1,73 @@
...
+
+            if sr == destination[0] and sc == destination[1]:
+                return True
+
+            for dr, dc in DIRS:
+                new_r, new_c = sr + dr, sc + dc
```
mujingw, line 63: `new_r` and `new_c` are okay, but they are not consistent with the style that you use for `sr` and `dr`. I would use `nr` and `nc` for consistency in style.

490.py
```diff
@@ -0,0 +1,73 @@
...
+
+                while new_r >= 0 and new_r < R and new_c >= 0 and new_c < C and maze[new_r][new_c] == 0:
+                    new_r += dr
+                    new_c += dc
+
+                if not visited[new_r - dr][new_c - dc]:
+                    queue.append([new_r - dr, new_c - dc])
+                    visited[new_r - dr][new_c - dc] = True
```
mujingw, line 71: Regarding this `visited` 2D array, see my comments [here](https://github.com/bunnygod/leetcode-python/pull/234#discussion_r679668416).

490.py
bunnygod: Fixed.

490.py
bunnygod: Sure, fixed.

PR #239, opened 2021-06-16
Check in 694.py.

bunnygod:


694.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return len(distinct_islands)
+
+    def bfs(self, grid: List[List[int]], row: int, col: int):
+        queue = deque([(row, col)])
+        island = []
+
+        while queue:
+            r, c = queue.popleft()
+            grid[r][c] = 0
```
mujingw, line 27: There is no need to have this line to mark `grid[r][c]` to `0`; line 41 marks this cell already. This only time this line is useful is for the first element in `queue`. Therefore, you can move this line to line 23 right after your `queue` initialization.

694.py
```diff
@@ -0,0 +1,44 @@
+class Solution:
```
mujingw, line 1: Overall this is a well written solution. This is my first time seeing a BFS solution for this classic LeetCode problem and I have learned something new today.

694.py
```diff
@@ -0,0 +1,45 @@
...
+    DIRS = {
+        (1, 0) : 'D',
+        (-1, 0): 'U',
```
mujingw, line 4: Do we really a dictionary here? A tuple of three elements `(dr, dc, dv)` would work well.

mujingw:
Sharing my DFS version below:

```python3
def numDistinctIslands(self, grid: List[List[int]]) -> int:
    def dfs(curr, r, c):
        if not (0 <= r < R and 0 <= c < C):
            return curr
        
        if grid[r][c] == 0:
            return curr
        
        if (r, c) in visited:
            return curr
        
        visited.add((r, c))
        
        for dx, dy, di in DIR:
            nr, nc = r + dx, c + dy
            curr += (dfs(di, nr, nc) + OPPOSITE[di])
        
        return curr
            
    R, C = len(grid), len(grid[0])
    DIR = ((0, 1, 'R'), (0, -1, 'L'), (-1, 0, 'U'), (1, 0, 'D'))
    OPPOSITE = {'R':'L', 'D':'U', 'U':'D', 'L':'R'}
    visited, islands = set(), set()
    
    for r, c in product(range(R), range(C)):
        if grid[r][c] == 1:
            signature = dfs('S', r, c)
            
            if signature != 'S':
                islands.add(signature)
    
    return len(islands)

PR #240, opened 2021-06-18
Check in 66.py.

bunnygod:


66.py
```diff
@@ -0,0 +1,18 @@
...
+
+        for i in range(len(digits) - 1, -1, -1):
+            digit = digits[i] + carry
+
+            if digit > 9:
+                digits[i] = 0
+                carry = 1
+            else:
+                digits[i] = digit
+                carry = 0
```
mujingw, line 13: You can enable early return here. If there is no carry at this digit, there won't be any carries afterwards.

66.py
```diff
@@ -0,0 +1,18 @@
+class Solution:
+    def plusOne(self, digits: List[int]) -> List[int]:
+        carry = 1
```
mujingw, line 3: Initializing `carry` to `0` makes more sense. It only becomes `1` when there is a transfer of digit `1` from those less significant digits.

Here is my version:

```python

def plusOne(self, digits: List[int]) -> List[int]:
    carry, addend = 0, 1
    
    for i in range(len(digits)-1, -1, -1):
        total = digits[i] + addend + carry
        digits[i] = total % 10
        carry = total // 10
        
        if carry == 1:
            addend = 0
        else:
            return digits
        
    return [1] + digits

PR #241, opened 2021-06-20
Check in 114.py.

bunnygod:


mujingw (commented):
This is a very interesting iterative approach and I never thought about solving this problem this way. My first impression after seeing your code was that it looked like the iterative pre-order traversal a lot! I then went back to this problem's page on LeetCode and suddenly noticed that it clearly says `The "linked list" should be in the same order as a pre-order traversal of the binary tree.` Aha!

mujingw:
Sharing my recursive approach here:

```python3
def flatten(self, root: Optional[TreeNode]) -> None:
    """
    Do not return anything, modify root in-place instead.
    """
    if not root:
        return None
    
    flattened_left_subtree = self.flatten(root.left)
    flattened_right_subtree = self.flatten(root.right)
    
    if flattened_left_subtree:
        tail_of_left = flattened_left_subtree
        
        while tail_of_left.right:
            tail_of_left = tail_of_left.right
        
        tail_of_left.right = root.right
        root.right = flattened_left_subtree
        root.left = None
    
    return root

**Comment by mujingw** [comment]
Refactored my code a little bit. This is a classic application of post-order traversal on binary trees. I extracted the post-order traversal logic into its own function to make this clear. Also, we were told not to return anything from the `flatten` function, so the previous implementation did not really follow this requirement.  

```python3
def flatten(self, root: Optional[TreeNode]) -> None:
    """
    Do not return anything, modify root in-place instead.
    """
    self.postorder(root)

def postorder(self, node):
    if not node:
        return None
    
    left = self.postorder(node.left)
    right = self.postorder(node.right)
    
    if left:
        tail = left
        
        while tail.right:
            tail = tail.right
            
        tail.right = node.right
        node.right = left
    
    node.left = None
    
    return node
```

PR #242, opened 2021-07-14
Check in 253.py.

bunnygod:


253.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
```
mujingw, line 1: For problems like this, I prefer using the "sweep line" technique, which is much easier to explain, code, and apply to many other problems. 

```python
def minMeetingRooms(self, intervals: List[List[int]]) -> int:
    d = defaultdict(int)

    for s, e in intervals:
        d[s] += 1
        d[e] -= 1

    return max(accumulate(d[t] for t in sorted(d.keys())))

```
The reasoning behind this is:

1. Regardless of what happens to all the other meetings, it is always true that each start of a meeting causes one _more_ meeting room to be occupied (if you look at each meeting individually and isolate itself from all the other meetings)
2. Similarly, each end of a meeting causes one meeting room to be released 
3. If we "sweep" each individual point along the timeline in a sorted order (i.e., from left to right on the X-axis), we will be able to keep track of the **total number** of meeting rooms occupied at any given point in time
4. The maximum such **total number** mentioned above is the minimum number of meeting rooms need to accommodate all meetings 

This exact technique can be conveniently applied to: [LC-252](https://leetcode.com/problems/meeting-rooms/), [LC-729](https://leetcode.com/problems/my-calendar-i/), [LC-731](https://leetcode.com/problems/my-calendar-ii/), [LC-732](https://leetcode.com/problems/my-calendar-iii/), [LC-759](https://leetcode.com/problems/employee-free-time/description/), [LC-986](https://leetcode.com/problems/interval-list-intersections/), [LC-1094](https://leetcode.com/problems/car-pooling/), [LC-1229](https://leetcode.com/problems/meeting-scheduler/), and [LC-1272](https://leetcode.com/problems/remove-interval/).

You can also refer to my previous explanation regarding this sweep line technique [here](https://github.com/bunnygod/leetcode-python/pull/156#issuecomment-835863378).

mujingw (commented):
LGTM

PR #243, opened 2021-08-05
Check in 1779.py.

bunnygod:


mujingw (approved):
LGTM. I came up with this one-liner just for fun:

```python3
def nearestValidPoint(self, x: int, y: int, points: List[List[int]]) -> int:
    return min([(abs(x - xi) + abs(y - yi), i) for i, (xi, yi) in enumerate(points) if xi == x or yi == y], default=((x, y), -1))[1]

PR #244, opened 2021-11-02
Check in 36.py.

bunnygod:


36.py
```diff
@@ -0,0 +1,29 @@
+class Solution:
+    def isValidSudoku(self, board: List[List[str]]) -> bool:
```
mujingw, line 2: I don't have any particular insight on this problem. I just wanted to share the following compact solution:

```python3
def isValidSudoku(self, board: List[List[str]]) -> bool:
    seen = []
    
    for r, row in enumerate(board):
        for c, ch in enumerate(row):
            if ch != '.':
                seen.append((ch, c))
                seen.append((r, ch)) 
                seen.append((r // 3, c // 3, ch))
    
    return len(seen) == len(set(seen))

mujingw: Hopefully it's easier to reproduce or explain during an interview due to its simplicity.

bunnygod: Genius solution!

PR #245, opened 2021-11-08
Check in 1971.py.

bunnygod:


1971.py
```diff
@@ -0,0 +1,20 @@
...
+            graph[e].append(s)
+
+        queue = deque([start])
+
+        while queue:
+            curr = queue.pop()
+
+            if curr == end:
+                return True
+            elif curr in graph and curr not in seen:
```
mujingw, line 16: Your solution may have been accepted by LeetCode, but I can tell that you are confused by the meaning and purpose of this `seen` variable.

In BFS, we use this `seen` variable to denote that a vertex has been added to the queue, such that the same vertex **does not get added more than once and avoids infinite loop**. Here in line 17, you are adding the neighbors of `curr` to the queue, not `curr` itself. You shouldn't be adding `curr` to `seen` here; instead, you should add every vertex in `graph[curr]` to `seen` because you just added them to the queue.

By the same reasoning, in line 10, you should add `start` to `seen` because that's where vertex `start` is seen for the first time and that's where it is added to the queue.

Note that we name this variable to `seen` not because it's cool or something. It literally means a vertex is seen, discovered, and just got added to the queue. It does not mean something like "the vertex has been visited or processed", which seems to be the case in your mind right now.

Regarding the different between `visited` and `seen`, I mentioned it a few months ago [here](https://github.com/bunnygod/leetcode-python/pull/237#discussion_r680713327).

mujingw (changes requested):

mujingw:
Also, try solving this problem using DFS and Union Find for more practice. During an interview, you are likely to be asked to compare the pros and cons of DFS, BFS, and Union Find. For this particular problem, since it's not shortest path-related, BFS is probably not the most efficient algorithm to be applied here, even though its time complexity is in the same category as DFS'.

mujingw:
Once you are done with this problem, try [LC-841](https://leetcode.com/problems/keys-and-rooms/) and see how well you have mastered this topic.

1971.py
```diff
@@ -0,0 +1,48 @@
+# class Solution:
```
mujingw, line 1: Manually commenting out any portion of your code in a version controlled environment is not a good practice. I would create a new file for a different implementation, or just overwrite the existing version with the new version. Ultimately, providing you access to older versions of your code is exactly what a version control system is supposed to do.

1971.py
```diff
@@ -0,0 +1,48 @@
...
+#             for v in graph[curr]:
+#                 if v not in seen:
+#                     queue.append(v)
+#                     seen.add(v)
+
+#         return False
+
+class Solution:
+    def validPath(self, n: int, edges: List[List[int]], start: int, end: int) -> bool:
+        seen, graph = set(), defaultdict(list)
```
mujingw, line 27: The logic of this piece of code is correct. However, it seems that you are still confusing yourself with `seen` and `visited`. 

In BFS, we maintain a set of nodes in the `seen` variable because these nodes have just been "seen" and not "visited" yet. 

Here in DFS, when you add the node to the set (in line 40), you have already "visited" that node (the current `dfs()` recursive call belongs to the `node` you passed in). Therefore, naming it to `visited` is more appropriate and should help you understand the subtle difference between DFS and BFS when it comes handling the "states" of the nodes.

1971.py
bunnygod: Instead, I listed all the solutions in one file. I don't need it to be compiled.

PR #246, opened 2021-11-12
Check in 841.py.

bunnygod:


841.py
```diff
@@ -0,0 +1,38 @@
...
+    def canVisitAllRooms(self, rooms: List[List[int]]) -> bool:
+        visited = set()
+
+        def dfs(room: int):
+            if room in visited:
+                return False
+
+            visited.add(room)
+
+            if len(visited) == len(rooms):
```
mujingw, line 11: When designing or writing recursive functions, we should always ask ourselves what the base case is (well, sometimes there are more than one base case) and what the recursive cases are.

Here, `len(visited) == len(rooms)` is a condition related to the `canVisitAllRooms` problem, not the problem this `dfs()` recursive function is supposed to solve. Essentially you are mixing up or coupling two problems into one function, which is not a good practice. And when the problem gets more complex than this simple DFS problem, you will get lost in the middle very easily.

841.py
```diff
@@ -0,0 +1,32 @@
+# DFS Solution
+class Solution:
+    def canVisitAllRooms(self, rooms: List[List[int]]) -> bool:
+        def dfs(room: int):
+            if room not in visited:
```
mujingw, line 5: This `dfs()` recursive function is missing a base case.

PR #247, opened 2021-11-13
Check in 261.py.

bunnygod:


261.py
```diff
@@ -0,0 +1,84 @@
...
+class Solution:
+    def validTree(self, n: int, edges: List[List[int]]) -> bool:
+        adjacency = defaultdict(list)
+
+        for a, b in edges:
+            adjacency[a].append(b)
+            adjacency[b].append(a)
+
+        stack = [0]
+        parent = {0: -1}
```
mujingw, line 14: Technically this `parent` variable does not represent parent-child relationships, because:

1. We don't know if the edges given in the input actually form a tree until the very end of this function, so at this point we simply don't know whether a tree structure or a parent-child relationship exists (what we do know right now are just a list of generic edges)
2. The starting vertex `0` is not necessarily the root of the _potential_ tree, so you are not necessarily going from the root node down to the child nodes for this variable named `parent` to make sense

This variable is more about saving the edges that you have explored so far.

bunnygod: I see your points. I named it `parent` to indicate that we'll store the other node on the other side of an edge into the collection of edges seen to avoid the undirected edge being checked twice.

PR #248, opened 2021-11-13
Check in 547.py.

bunnygod:


547.py
```diff
@@ -0,0 +1,46 @@
+# This problem is to find number of connected component in an un-directed graph.
+
+# BFS Iterative Solution
+class Solution:
+    def findCircleNum(self, isConnected: List[List[int]]) -> int:
+        N = len(isConnected)
+        visited = [0] * N
```
mujingw, line 7: Using `0` for false and `1` for true is very much a C-style programming. In Python we have `True` and `False` keywords so it's better to use them for clarity.

547.py
```diff
@@ -0,0 +1,46 @@
...
+        for i in range(N):
+            if not visited[i]:
+                queue.append(i)
+
+                while queue:
+                    node = queue.popleft()
+                    visited[node] = 1
+
+                    for j in range(N):
+                        if isConnected[node][j] == 1 and visited[j] == 0:
```
mujingw, line 20: Maybe for this particular problem it happens to be right. But if you use this template to solve [LC-200](https://leetcode.com/problems/number-of-islands/) you would probably get a TLE due to `j` being added to `queue` repeatedly.

mujingw, line 20: Maybe for this particular problem it happens to be right. But if you use this template to solve [LC-200](https://leetcode.com/problems/number-of-islands/) you would probably get a TLE due to `j` being added to `queue` repeatedly.

547.py
```diff
@@ -0,0 +1,46 @@
...
+                            queue.append(j)
+
+                count += 1
+
+        return count
+
+# DFS Recursive Solution
+class Solution:
+    def dfs(self, isConnected: List[List[int]], visited: List[int], i: int) -> None:
+        for j in range(len(isConnected)):
```
mujingw, line 30: For recursive functions like this `dfs()` one, I would expect to see a pattern in your code that clearly separates the base case and the recursive cases (but I don't see one here). Your code happens to be right because some of the base cases are trivial and the implicit `return` statement added at the end of the function helps you address the (trivial) base case.

In general, I don't see that you have developed your own habit or template to implement BFS and DFS algorithms. Each solution seems to be an ad-hoc, one-off implementation that usually only works for that particular problem.

bunnygod: Got it. Any suggestion on how to separate the base case and recursive case? How is the new revision look like?

PR #249, opened 2021-11-13
Check in 323.py.

bunnygod:


323.py
```diff
@@ -0,0 +1,82 @@
+# BFS Soluting using Adjacency Matrix.
```
mujingw, line 1: My comments to the three implementations here would be the same as mentioned in [LC-547](https://github.com/bunnygod/leetcode-python/pull/248#pullrequestreview-806270573).

I will come back to this PR tomorrow with my own BFS and DFS implementations, so that you get a better view of what I am saying.

bunnygod: Any updates?

mujingw: In [LC-1723](https://github.com/bunnygod/leetcode-python/pull/253/commits/ed55476cd54e5fc836926de5e11ce0f2177f7f3b) you have demonstrated your most recent understanding of writing correct and well maintained BFS code, so I won't be providing my BFS version here anymore. 

For DFS, this is what I just submitted:

```python3
def countComponents(self, n: int, edges: List[List[int]]) -> int:
    self.g = defaultdict(set)
    
    for u, v in edges:
        self.g[u].add(v)
        self.g[v].add(u)
        
    self.visited = set()
    res = 0
    
    for i in range(n):
        if i not in self.visited:
            self.dfs(i)
            res += 1
            
    return res
    
def dfs(self, node):
    if node in self.visited:
        return 
    
    self.visited.add(node)
    
    for ng in self.g[node]:
        self.dfs(ng)
 ```

This problem is a twin brother of [LC-200](https://leetcode.com/problems/number-of-islands/), and it's even easier than LC-200 due to its simpler boundary conditions.

bunnygod: It is still very hard for me to code a nit solution based on DFS. Also, I am still not 100% clear in which case I should use `visited` and in which case I should use `seen`. To be clear that, I know how to name them based on where we populate the collection, I just don't know when to populate them is the best.

PR #251, opened 2021-11-16
Check in 210.py.

bunnygod:


mujingw (commented):
LGTM. Essentially these ([LC-207](https://github.com/bunnygod/leetcode-python/pull/250#pullrequestreview-807338249), LC-210, [LC-1136](https://github.com/bunnygod/leetcode-python/pull/252#pullrequestreview-807351042)) are the same problem.

PR #254, opened 2021-11-22
Check in 133.py.

bunnygod:


Graph/Basic Graph Traversal/133.py
```diff
@@ -0,0 +1,49 @@
...
+        self.neighbors = neighbors if neighbors is not None else []
+"""
+
+# BFS Solution
+class Solution:
+    def cloneGraph(self, node: 'Node') -> 'Node':
+        if not node:
+            return node
+
+        graph = {}
```
mujingw, line 15: The name `graph` is misleading here. It's not a representation of any graph; it's a mapping between those original vertices and their corresponding cloned ones.

Graph/Basic Graph Traversal/133.py
```diff
@@ -0,0 +1,49 @@
...
+        queue = deque([node])
+        graph[node] = Node(node.val, [])
+
+        while queue:
+            n = queue.popleft()
+
+            for neighbor in n.neighbors:
+                if neighbor not in graph:
```
mujingw, line 23: Do you recognize the **_role_** this `graph` variable plays here (other than storing the cloned vertices)?

bunnygod: It works as `seen`.

mujingw: Great! This is exactly what I wanted to hear.

PR #255, opened 2021-11-24
Check in 1462.py.

bunnygod:


Graph/Basic Graph Traversal/1462.py
```diff
@@ -0,0 +1,55 @@
...
+
+            for v in self.graph[curr]:
+                if v not in seen:
+                    queue.append(v)
+                    seen.add(v)
+
+        return False
+
+    # DFS Solution
+    # Time Limited Exceeded.
```
mujingw, line 34: This solution gets TLE not because DFS (plus memoization) is inefficient. Rather, it is because there is a programming error in your code. I won't disclose too much for now; let's see if you can find this error yourself.

Graph/Basic Graph Traversal/1462.py
```diff
@@ -0,0 +1,55 @@
...
+        for a, b in prerequisites:
+            self.graph[a].append(b)
+
+        for u, v in queries:
+            ans.append(self.isPrerequisites(u, v))
+
+        return ans
+
+    # BFS Solution.
+    def isPrerequisites(self, start: int, end: int) -> bool:
```
mujingw, line 16: This BFS implementation is good! Hopefully now you can compare and contrast these three types of BFS problems:

1. Grids
2. Trees
3. Implicit graph problems modeled into a graph (in the form of adjacency list or adjacency matrix)

Graph/Basic Graph Traversal/1462.py
```diff
@@ -0,0 +1,55 @@
...
+        while queue:
+            curr = queue.popleft()
+
+            if curr == end:
+                return True
+
+            for v in self.graph[curr]:
+                if v not in seen:
+                    queue.append(v)
+                    seen.add(v)
```
mujingw, line 29: This `seen` variable is very important in BFS implementations. You are doing it correctly here.

Graph/Basic Graph Traversal/1462.py
bunnygod: I should recursively call function `dfs` instead of `isPrerequisites`.

However, both BFS/DFS solutions do not show a competitive time/space complexity in LC. Not sure if that is because the nature of the problem and the nature of Python, or there is more efficient way to do this problem.

PR #263, opened 2021-12-01
Check in 57.py.

bunnygod:


mujingw (commented):
Good to see you use the sweep line technique here. I treated this problem almost the same as [LC-56](https://leetcode.com/problems/merge-intervals/) and here is my solution:

```python3
def insert(self, intervals: List[List[int]], new_interval: List[int]) -> List[List[int]]:
    all_intervals = intervals + [new_interval]
    all_intervals.sort()
    res = [[all_intervals[0][0], all_intervals[0][1]]]
    
    for start, end in all_intervals[1:]:
        if start > res[-1][1]:
            res.append([start, end])
        else:
            res[-1][1] = max(res[-1][1], end)
    
    return res
```

In fact, this problem can be decomposed as this:

```python3
def insert(intervals, new_interval):
    intervals.append(new_interval)

    return merge(intervals)
``` 
The `merge()` function is what you would submit for LC-56.

